PR #24, opened 2025-04-14
Add signed integer support with a multiplication stub; add unit tests

cedricjiang:


parasol_runtime/src/circuits/mul.rs
```diff
@@ -12,6 +12,15 @@ use crate::{
     fluent::Muxable,
 };
 
+/// Compute the product of 2 N-bit unsigned values a and b.
+pub fn append_int_multiply<OutCt: Muxable>(
+    uop_graph: &mut FheCircuit,
+    a: &[NodeIndex],
+    b: &[NodeIndex],
+) -> (Vec<NodeIndex>, Vec<NodeIndex>) {
+    todo!("append_int_multiply ({:#?}, {:#?}, {:#?})", uop_graph, a, b);
```
cedricjiang, line 21: dummy...to be implemented...the message is to make compiler not to complain about the unused variables

parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -20,8 +20,11 @@ use petgraph::stable_graph::NodeIndex;
 use serde::{Deserialize, Serialize};
 use sunscreen_tfhe::entities::Polynomial;
 
-/// Marker trait for distinguishing unsigned and signed integer types
-pub trait Sign {}
+/// Trait for distinguishing unsigned and signed integer types
+pub trait Sign {
+    /// Compare circuit generation function for this sign
+    fn gen_compare_circuit(max_len: usize, gt: bool, eq: bool) -> MuxCircuit;
```
cedricjiang, line 26: moved the function to associate with this trait so the generic test works

cedricjiang: may need to do this for functions that are individually implemented, that is `resize` and `mul`

parasol_runtime/src/fluent/int.rs
```diff
@@ -0,0 +1,193 @@
+use crate::{L1GgswCiphertext, circuits::mul::append_int_multiply};
```
cedricjiang, line 1: pretty much a copy of unsigned :)

parasol_runtime/src/fluent/int.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// If M < N, this will truncate the high-order bits with sign bit preserved.
+    /// If M == N, why did you call this? In any case, the returned nodes will equal the input nodes.
+    ///
+    /// This operation is "free" in that it adds no computation to the graph.
+    pub fn resize<const M: usize>(&self, ctx: &'a FheCircuitCtx) -> IntGraphNodes<'a, M, T> {
+        let extend = 1 + if M > N { M - N } else { 0 };
+
+        let min_len = M.min(N) - 1;
+
+        let sign_bit = self.bits.last().unwrap();
```
cedricjiang, line 37: copy sign bit for extension - is this correct? I assume ok since it's just an encrypted version of the sign bit

cedricjiang: if not, I don't see a way to achieve signed extension

parasol_runtime/src/fluent/int.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub fn mul<OutCt: Muxable>(
+        &self,
+        other: &Self,
+        ctx: &'a FheCircuitCtx,
+    ) -> IntGraphNodes<'a, N, OutCt> {
+        let a = self.bits.iter().map(|x| x.node).collect::<Vec<_>>();
+
+        let b = other.bits.iter().map(|x| x.node).collect::<Vec<_>>();
+
+        let (lo, _hi) = append_int_multiply::<OutCt>(&mut ctx.circuit.borrow_mut(), &a, &b);
```
cedricjiang, line 65: call the signed version (to be implemented)

parasol_runtime/src/fluent/mod.rs
```diff
@@ -360,6 +362,7 @@ impl Muxable for L1GlevCiphertext {
 #[cfg(test)]
 mod tests {
```
cedricjiang, line 363: we can discuss how to deal with the tests here, including moving

parasol_runtime/src/fluent/mod.rs
```diff
@@ -389,30 +392,37 @@ mod tests {
...
 
     #[test]
     fn can_roundtrip_l1_ggsw() {
-        roundtrip::<L1GgswCiphertext>();
+        roundtrip::<L1GgswCiphertext, Unsigned>();
+        roundtrip::<L1GgswCiphertext, Signed>();
     }
 
-    fn input_output<T: CiphertextOps>() {
+    fn input_output<T: CiphertextOps, U: Sign>(test_val: u64) {
```
cedricjiang, line 417: user to provide val so the signed version can use "negative" number

cedricjiang: same for a few below

parasol_runtime/src/fluent/mod.rs
```diff
@@ -630,12 +668,17 @@ mod tests {
...
+        case::<L1GlweCiphertext, Signed>(false, (65501, 65501));
+        case::<L1GlweCiphertext, Unsigned>(true, (43, 42));
+        case::<L1GlweCiphertext, Signed>(true, (65501, 65500));
+        case::<L1GlevCiphertext, Unsigned>(false, (43, 43));
+        case::<L1GlevCiphertext, Signed>(false, (65501, 65501));
+        case::<L1GlevCiphertext, Unsigned>(true, (43, 42));
+        case::<L1GlevCiphertext, Signed>(true, (65501, 65500));
     }
 
+    // TODO for signed
```
cedricjiang, line 681: I don't know how to do these correctly yet for signed. My understanding is the underlying code only deal with matched size, so before going there some code will take care of the unmatched size (essentially "resize"). I need to think a bit more about how to make that code do the correct resizing. May or may not do this in this PR

cedricjiang: same for other "size mismatch" tests

parasol_runtime/src/fluent/mod.rs
```diff
@@ -762,16 +806,25 @@ mod tests {
         }
 
         fn cases<OutCt: Muxable>() {
-            case::<OutCt>(false, false);
-            case::<OutCt>(false, true);
-            case::<OutCt>(true, false);
-            case::<OutCt>(true, true);
+            case::<OutCt, Unsigned>(false, false, (43, 42));
+            case::<OutCt, Signed>(false, false, (65501, 65500));
+            // case::<OutCt, Signed>(false, false, (1, 65535));
```
cedricjiang, line 811: this one is commented out because it fails...need to understand why

parasol_runtime/src/fluent/mod.rs
```diff
@@ -762,16 +806,25 @@ mod tests {
...
+            case::<OutCt, Unsigned>(true, true, (43, 42));
+            case::<OutCt, Signed>(true, true, (65501, 65500));
+            // case::<OutCt, Signed>(true, true, (1, 65535));
         }
 
         cases::<L1GlweCiphertext>();
         cases::<L1GlevCiphertext>();
     }
 
+    // TODO for signed
```
cedricjiang, line 827: this test function has only one test cases so there's no inner "case" function to generalize, but easy to refactor. I will likely include include it in this PR

cedricjiang: same for the "plain" version below

cedricjiang:
I am bothered by that `cargo test` succeeded locally (however when I `cargo clean` and then `cargo test` it failed)

cedricjiang:
Hmm...tests seem flaky now

parasol_runtime/src/fluent/int.rs
rickwebiii: Yeah, you want to do a signed extension.

rickwebiii: Truncation is just that. If you lose the sign bit and now get a positive number, then so be it.

parasol_runtime/src/fluent/int.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// # Remarks
+    /// If M > N, this will sign extend the integer.
+    /// If M < N, this will truncate the high-order bits with sign bit preserved.
+    /// If M == N, why did you call this? In any case, the returned nodes will equal the input nodes.
+    ///
+    /// This operation is "free" in that it adds no computation to the graph.
+    pub fn resize<const M: usize>(&self, ctx: &'a FheCircuitCtx) -> IntGraphNodes<'a, M, T> {
+        let extend = 1 + if M > N { M - N } else { 0 };
+
+        let min_len = M.min(N) - 1;
```
rickwebiii, line 35: I think this should not be -1.

parasol_runtime/src/fluent/int.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// Convert this `N`-bit integer to an `M`-bit integer of the same ciphertext type.
+    ///
+    /// # Remarks
+    /// If M > N, this will sign extend the integer.
+    /// If M < N, this will truncate the high-order bits with sign bit preserved.
+    /// If M == N, why did you call this? In any case, the returned nodes will equal the input nodes.
+    ///
+    /// This operation is "free" in that it adds no computation to the graph.
+    pub fn resize<const M: usize>(&self, ctx: &'a FheCircuitCtx) -> IntGraphNodes<'a, M, T> {
+        let extend = 1 + if M > N { M - N } else { 0 };
```
rickwebiii, line 33: I think there should not be a +1 here.

parasol_runtime/src/fluent/mod.rs
```diff
@@ -516,29 +537,43 @@ mod tests {
...
+        case::<L1GlweCiphertext, Signed>(false, true, (1, 65535));
+        case::<L1GlweCiphertext, Unsigned>(true, false, (43, 42));
+        case::<L1GlweCiphertext, Signed>(true, false, (65501, 65500));
+        case::<L1GlweCiphertext, Signed>(true, false, (1, 65535));
+        case::<L1GlweCiphertext, Unsigned>(true, true, (43, 42));
+        case::<L1GlweCiphertext, Signed>(true, true, (65501, 65500));
+        case::<L1GlweCiphertext, Signed>(true, true, (1, 65535));
+
+        case::<L1GlevCiphertext, Unsigned>(false, false, (43, 42));
+        // case::<L1GlevCiphertext, Signed>(false, false, (65501, 65500));
```
rickwebiii, line 554: Remove commented lines

rickwebiii (approved):

parasol_runtime/src/fluent/int.rs
cedricjiang: if we preserve the sign bit then we +1 here, the reason for that is the min_len has -1 so the sign bit is not included anyway in either truncation or extension, added a comment.

parasol_runtime/src/fluent/int.rs
cedricjiang: by doing so (preserving the sign bit in truncation) the code is more consistent as you don't need to handle truncation and extension differently

PR #28, opened 2025-04-16
Basic memory allocation from ELF segments + mmap

rickwebiii:


parasol_cpu/src/error.rs
```diff
@@ -81,14 +81,8 @@ pub enum Error {
...
     /// A load or store occured out of a memory's bounds.
-    #[error("(inst:{inst_id}, pc:0x{pc:x}) Attempted to access data out of bounds")]
-    AccessViolation {
-        /// The faulting instruction's id.
-        inst_id: usize,
-
-        /// The program counter at time of error.
-        pc: usize,
-    },
+    #[error("Attempted to access address 0x{0:8x}, which was not mapped.")]
```
cedricjiang, line 84: could just say "access unmapped address", conciser

parasol_cpu/src/error.rs
```diff
@@ -179,13 +173,30 @@ pub enum Error {
     #[error("The ELF file does not contain the specified symbol: {0}")]
     ElfSymbolNotFound(String),
 
+    /// When parsing the ELF file, encountered an out-of-bounds file offset.
+    #[error("The given ELF byte offset {0} exceeds the file's length.")]
+    ElfByteOutofBounds(u32),
```
cedricjiang, line 178: capital O for of?

parasol_cpu/src/error.rs
```diff
@@ -179,13 +173,30 @@ pub enum Error {
...
+
     /// Attempted to allocate a virtual address that's already mapped.
     #[error("Failed to allocate virtual address space. Already in use.")]
     VirtualAddressInUse,
 
     /// Failed to create a [`std::ffi::CString`]`
     #[error("Failed to create CString: {0}")]
     CStringCreationError(#[from] std::ffi::NulError),
+
+    /// Cannot fulfil the given mmap request as no contiguous address region exists of the
```
cedricjiang, line 188: I think it's fulfill

parasol_cpu/src/memory/mod.rs
```diff
@@ -51,47 +76,146 @@ impl Memory {
...
-
-            //memory.try_allocate_at(vaddr, len)?;
+            let mem_start = Ptr32::from(segment.p_vaddr as u32);
+            let mem_end = mem_start.try_offset(segment.p_memsz as u32)?;
+
+            // Not actually a pointer, but ELF32 specifies 32-bit offsets for files.
+            let file_start = Ptr32::from(segment.p_offset as u32);
+            let file_end = file_start.try_offset(segment.p_filesz as u32)?;
+
+            // TODO: WTF do we do with zero-size segments?
```
cedricjiang, line 96: are zero size segments legal at all

parasol_cpu/src/memory/mod.rs
```diff
@@ -51,47 +76,146 @@ impl Memory {
...
         let mut pages = self.pages.lock().unwrap();
 
-        // Make sure none of the pages are in use.
-        if (start_page_id..=end_page_id).any(|i| pages[i].is_some()) {
-            return Err(Error::VirtualAddressInUse);
-        }
-
         for i in start_page_id..=end_page_id {
-            pages[i] = Some(Page::allocate())
+            if pages[i].is_none() {
```
cedricjiang, line 140: can be done with [get_or_insert](https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert) I think?

cedricjiang: we don't need the get part though, it's like "insert if none"

parasol_cpu/src/memory/mod.rs
```diff
@@ -51,47 +76,146 @@ impl Memory {
...
+        let page_id = Page::page_id_from_pointer(virtual_address) as usize;
+
+        let mut pages = self.pages.lock().unwrap();
+
+        match &mut pages[page_id] {
+            Some(p) => {
+                let page_offset = Page::offset_from_pointer(virtual_address);
+
+                {
+                    p.store_byte(page_offset as usize, data);
```
cedricjiang, line 173: why this is in a block with itself?

parasol_cpu/src/memory/mod.rs
```diff
@@ -116,19 +240,128 @@ impl Page {
...
+        let mask = PAGE_SIZE - 1;
+
+        assert_eq!(offset, addr & mask);
+    }
+
+    #[test]
+    fn page_id() {
+        let addr = 0x12345678;
+        let mask = PAGE_SIZE - 1;
+        let mask = !mask;
```
cedricjiang, line 361: why not just `let mask = !(PAGE_SIZE - 1);`

parasol_cpu/src/error.rs
```diff
@@ -81,14 +81,8 @@ pub enum Error {
...
-    #[error("(inst:{inst_id}, pc:0x{pc:x}) Attempted to access data out of bounds")]
-    AccessViolation {
-        /// The faulting instruction's id.
-        inst_id: usize,
-
-        /// The program counter at time of error.
-        pc: usize,
-    },
+    #[error("Attempted to access address 0x{0:8x}, which was not mapped.")]
+    AccessViolation(u32),
```
cedricjiang, line 85: this now no longer carries the information it used to

parasol_cpu/src/error.rs
rickwebiii: Yeah, one `l` is the British spelling, which I guess is why my spell checker didn't flag.

parasol_cpu/src/memory/mod.rs
rickwebiii: Our linker did output one for the stack, so I guess they're legal in ELF.

parasol_cpu/src/error.rs
rickwebiii: Correct, as memory operations can occur outside an FHE program (i.e. when the users is populating memory with their data).

cedricjiang: just saying, but if the program counter etc information is available, it seems helpful to also carry them - they could be `Option` I guess? not a strong opinion

parasol_cpu/src/memory/mod.rs
cedricjiang: oh I see, it means the file has this section as "zero size" but it says in the memory it'll have a non zero size?

PR #29, opened 2025-04-17
Make more unit tests work on both signed and unsigned int with required refactoring

cedricjiang:


parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -33,6 +33,14 @@ pub trait Sign {
         a: &[NodeIndex],
         b: &[NodeIndex],
     ) -> (Vec<NodeIndex>, Vec<NodeIndex>);
+
+    /// Resize implementation function for this sign
+    fn resize<T: CiphertextOps>(
```
cedricjiang, line 38: this extracts "essential" components that are not relevant to the "context" so it can be implemented by sign

parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -33,6 +33,14 @@ pub trait Sign {
...
+        input: &[BitNode<T>],
+        zero: &BitNode<T>,
```
cedricjiang, line 40: for zero extension, you cannot identify "zero" from the input and creating zero requires the "context", not used by signed version

parasol_runtime/src/fluent/mod.rs
```diff
@@ -824,59 +827,73 @@ mod tests {
...
-        let sel_false = Bit::<L1GgswCiphertext>::encrypt_secret(false, enc, &sk).graph_input(&ctx);
-        let sel_true = Bit::<L1GgswCiphertext>::encrypt_secret(true, enc, &sk).graph_input(&ctx);
+            let sel_false =
+                Bit::<L1GgswCiphertext>::encrypt_secret(false, enc, &sk).graph_input(&ctx);
+            let sel_true =
+                Bit::<L1GgswCiphertext>::encrypt_secret(true, enc, &sk).graph_input(&ctx);
 
-        let a = UInt::<16, L1GlweCiphertext>::encrypt_secret(42, enc, &sk).graph_inputs(&ctx);
-        let b = UInt::<16, L1GlweCiphertext>::encrypt_secret(24, enc, &sk).graph_inputs(&ctx);
+            let a = GenericInt::<16, L1GlweCiphertext, U>::encrypt_secret(test_vals.0, enc, &sk)
```
cedricjiang, line 843: note the use of generic int

cedricjiang: same below for the "plain" one

parasol_runtime/src/fluent/generic_int.rs
rickwebiii: I think you should have a method on GenralInt that takes an `extend_bit`. Then, for uint, you pass zero, for int you pass MSB. The rest of the function is the same is common to both.

rickwebiii (approved):

parasol_runtime/src/fluent/generic_int.rs
cedricjiang: there are a few issues, one is there needs to be a `resize` method on `GenericIntGraphNodes` for the test to work on both signed and unsigned (whether we need to unify the two types in testing is arguable but let's say we go with that). as a result, there must be some methods in the `Sign` trait to provide the help to distinguish the differences. another thing is that I decide to make signed truncation keep the sign bit (which is desirable in my opinion) so the actual implementation differ by that "+1 / -1" thing.

cedricjiang: that said, the method in `Sign` trait essentially could just return a "resize config" comprised on `min_len`, `extend` and `bit_to_extend`

PR #30, opened 2025-04-17
Support arithmetic shift in the processor

cedricjiang:


mux_circuits/src/bitshift.rs
```diff
@@ -45,6 +46,10 @@ pub fn bitshift(inputs: u16, shift_size: u16, right: bool, zeros: bool) -> MuxCi
         panic!("Shift without zeros is only supported for power of two inputs.");
     }
 
+    if !zeros && arith {
+        panic!("Arithmetic shift is not compatible with rotation");
```
cedricjiang, line 50: dunno...but fair I think

mux_circuits/src/bitshift.rs
```diff
@@ -58,6 +63,12 @@ pub fn bitshift(inputs: u16, shift_size: u16, right: bool, zeros: bool) -> MuxCi
         .map(|i| variable_set.mk_var(vars[i as usize]))
         .collect::<Vec<_>>();
 
+    let old_msb = if arith {
+        Some(result.last().unwrap().clone())
```
cedricjiang, line 67: don't clone if we are not doing arithmetic shift

mux_circuits/src/bitshift.rs
```diff
@@ -112,6 +123,10 @@ pub fn bitshift(inputs: u16, shift_size: u16, right: bool, zeros: bool) -> MuxCi
         }
     }
 
+    if arith {
+        *result.last_mut().unwrap() = old_msb.unwrap();
```
cedricjiang, line 127: unwrap is safe

cedricjiang: (I mean for `old_msb`)

mux_circuits/src/bitshift.rs
```diff
@@ -152,15 +168,18 @@ mod tests {
...
 
-        let circuit = bitshift(width, width, right, zeros);
+        let circuit = bitshift(width, width, right, zeros, arith);
 
         let mut expected_bits = (0..width)
             .rev()
             .map(|i| Bit((value >> i) & 0x1 == 1))
             .collect::<Vec<_>>();
 
+        let old_msb = *expected_bits.last().unwrap();
```
cedricjiang, line 181: for test code, I am not bothered (and this is just `Bit`)

mux_circuits/src/bitshift.rs
```diff
@@ -255,29 +280,37 @@ mod tests {
...
-            for right in [false, true] {
-                for width in 1..=6u16 {
-                    let mask = (1 << width) - 1;
-
-                    // Skip the cases where we would need to perform a modulus
-                    // operation.
-                    if !zeros && !(width.is_power_of_two()) {
-                        continue;
-                    }
+            for arith in [false, true] {
```
cedricjiang, line 283: add a layer

parasol_cpu/src/proc/ops/bitshift.rs
```diff
@@ -179,7 +179,7 @@ impl FheProcessor {
             pc,
             |val, shift, _| val >> shift,
             |c, shift, l1glwe_zero| encrypted_value_plain_shift(c, shift, l1glwe_zero, true, true),
-            |inputs, shift_size| bitshift(inputs as u16, shift_size as u16, true, true),
+            |inputs, shift_size| bitshift(inputs as u16, shift_size as u16, true, true, false),
```
cedricjiang, line 182: false arith is default for existing, the new ones have true

mux_circuits/src/bitshift.rs
```diff
@@ -45,6 +46,10 @@ pub fn bitshift(inputs: u16, shift_size: u16, right: bool, zeros: bool) -> MuxCi
         panic!("Shift without zeros is only supported for power of two inputs.");
     }
 
+    if (zeros || !right) && arith {
```
cedricjiang, line 49: arith only allowed when zeros is false and right is true

mux_circuits/src/bitshift.rs
```diff
@@ -58,6 +63,8 @@ pub fn bitshift(inputs: u16, shift_size: u16, right: bool, zeros: bool) -> MuxCi
         .map(|i| variable_set.mk_var(vars[i as usize]))
         .collect::<Vec<_>>();
 
+    let old_msb = if arith { Some(result[0].clone()) } else { None };
```
cedricjiang, line 66: this avoids cloning if not arith, not sure if it's needed as it makes some code a bit nasty

mux_circuits/src/bitshift.rs
```diff
@@ -91,7 +98,13 @@ pub fn bitshift(inputs: u16, shift_size: u16, right: bool, zeros: bool) -> MuxCi
                         intermediate[input_index].clone()
                     }
                 }
-                (true, false) => intermediate[input_index].clone(),
+                (true, false) => {
+                    if arith && input_index < shift {
+                        old_msb.as_ref().unwrap().to_owned()
```
cedricjiang, line 103: this is essentially the key to the method. arith can only be true in this match arm, and if that's the case, similar to filling `0` in the (true, true) arm, we fill in the sign bit

mux_circuits/src/bitshift.rs
```diff
@@ -100,15 +113,21 @@ pub fn bitshift(inputs: u16, shift_size: u16, right: bool, zeros: bool) -> MuxCi
     }
 
     // Now mux in the higher order wrapped select bits; if any are 1 then the
-    // entire output vector is zero.
-    if zeros {
+    // entire output vector is zero or sign
+    if zeros || arith {
```
cedricjiang, line 117: similarly, fill the entire thing if needed

parasol_cpu/src/proc/ops/bitshift.rs
```diff
@@ -179,7 +179,29 @@ impl FheProcessor {
...
+        self.shift_operation(
+            retirement_info,
+            dst,
+            src,
+            shift,
+            instruction_id,
+            pc,
+            |val, shift, _| val >> shift,
+            |c, shift, l1glwe_zero| encrypted_value_plain_shift(c, shift, l1glwe_zero, true, true),
+            |inputs, shift_size| bitshift(inputs as u16, shift_size as u16, true, false, true),
```
cedricjiang, line 204: only arith = true call with other two as permitted

cedricjiang: We have tests for the MUX circuit, but do we have tests for this `shra` function?

mux_circuits/src/bitshift.rs
ryanorendorff: At this point the inputs are annoying enough that it might be a good idea to change the calling structure. Mostly I feel like this code will be annoying (due to boolean blindness)

```rust
bitshift(inputs, size, true, false, true) // I have no idea what these bools mean
```

mux_circuits/src/bitshift.rs
ryanorendorff: Ah I see what is happening here. The clone should be cheap since it is part of the circuit definition and not adding into the circuit itself, so if there is a way around this use of `Option` it would be preferred.

Also if you could add a comment to what this for loop is doing it would be greatly appreciated!

ryanorendorff (approved):

parasol_cpu/src/proc/ops/bitshift.rs
ryanorendorff: Do you mean these? https://github.com/Sunscreen-tech/hpu/blob/main/parasol_cpu/src/proc/tests/bitshift.rs

mux_circuits/src/bitshift.rs
cedricjiang: I know...I thought about creating another function to just do this (which will only need `inputs` and `size`). actually @rickwebiii also wants that, however I realize that'll duplicate a lot of code (arithmetic shift is similar enough - just to copy the sign bit in every time essentially) so I added one more boolean.

Ideally we should use enums, note that there are only three choices for the "mode" that is not defined by two booleans which expose 4 options, the reason is "arith" shadows "zeros"

I probably should do this

mux_circuits/src/bitshift.rs
cedricjiang: yeah my idea is if you don't have arith true then you don't need to actually clone anything in the first place to create the value inside the "Some" version of `old_msb`. If it's cheap then I should avoid that, it does make the code a bit cleaner...

parasol_cpu/src/proc/ops/bitshift.rs
cedricjiang: oh yeah...need to add

mux_circuits/src/bitshift.rs
ryanorendorff: Do the enum! 😄

mux_circuits/src/bitshift.rs
```diff
@@ -23,7 +36,8 @@ fn ilog2_rounded_up(x: u16) -> u32 {
 ///   be specified MSB first.
 /// * `right` - Whether to shift right or left.
 /// * `zeros` - Whether to fill the shifted bits with zeros.
-pub fn bitshift(inputs: u16, shift_size: u16, right: bool, zeros: bool) -> MuxCircuit {
+/// * `arith` - Whether to do arithmetic shift (consider sign bit), requires right shift and not zeros.
+pub fn bitshift(inputs: u16, shift_size: u16, right: bool, mode: ShiftMode) -> MuxCircuit {
```
ryanorendorff, line 40: Make the shift direction an enum as well so that we don't have boolean blindness in the direction either.

ryanorendorff (approved):

mux_circuits/src/bitshift.rs
```diff
@@ -41,10 +55,14 @@ pub fn bitshift(inputs: u16, shift_size: u16, right: bool, zeros: bool) -> MuxCi
 
     // To handle this case we would need a modulus circuit, which is
     // non-trivial to implement.
-    if !zeros && !inputs.is_power_of_two() {
+    if mode != ShiftMode::Logical && !inputs.is_power_of_two() {
```
rickwebiii, line 58: Right shift for non-power-of two inputs should work just fine.

cedricjiang: oh yeah, rotation requires something complicated but zero fill vs sign fill should both work

rickwebiii (approved):

PR #32, opened 2025-04-22
Remove unused cruft

rickwebiii:


parasol_cpu/src/proc/mod.rs
```diff
@@ -507,7 +492,7 @@ impl Tomasulo for FheProcessor {
     ) -> Result<usize> {
         match dispatched_op {
             DispatchIsaOp::BranchNonZero(cond, target) => {
-                unwrap_registers!([self.constant_pool](cond));
+                unwrap_registers!((cond));
```
cedricjiang, line 495: do you need `((`?

rickwebiii: In this case, yeah, the macro matches on ()

parasol_cpu/src/proc/ops/mod.rs
```diff
@@ -164,17 +164,17 @@ pub fn make_parent_op(retirement_info: &RetirementInfo<DispatchIsaOp>) -> Arc<Co
...
-        let $reg = $const_pool.register(&$reg);
-        let $reg = std::ops::Deref::deref(&$reg);
-        unwrap_registers!([$const_pool] $($rest)*);
+    (($reg:ident) $($rest:tt)*) => {
+        let $reg = $reg.entry();
+        let $reg = std::ops::Deref::deref(&*$reg);
+        unwrap_registers!($($rest)*);
     };
-    ([$const_pool:expr]) => {
+    () => {
```
cedricjiang, line 178: I am bit surprised that rustfmt does not collapse this

cedricjiang: maybe because it's macro

parasol_cpu/src/proc/ops/store.rs
```diff
@@ -27,9 +27,10 @@ impl FheProcessor {
         pc: usize,
     ) {
         let store_impl = |scoreboard_entry: &ScoreboardEntryRef<DispatchIsaOp>| -> Result<()> {
-            let mut dst = self.constant_pool.register_force_mut(&dst)?;
-            let dst = dst.deref_mut();
-            unwrap_registers!([self.constant_pool](src));
+            let mut dst = dst.entry_force_mut();
+            let dst: &mut PtrRegister = dst.deref_mut();
```
cedricjiang, line 31: no inferrence now?

parasol_cpu/src/proc/mod.rs
rickwebiii: If you change the macro to not match on (), then you get uses that look like:

```rust
unwrap_registers!(mut dst a b);
```

which is hard to group the mut with the identifier at a glance.

cedricjiang: yeah makes sense

parasol_cpu/src/proc/ops/store.rs
```diff
@@ -27,14 +25,15 @@ impl FheProcessor {
         pc: usize,
     ) {
         let store_impl = |scoreboard_entry: &ScoreboardEntryRef<DispatchIsaOp>| -> Result<()> {
-            let mut dst = self.constant_pool.register_force_mut(&dst)?;
-            let dst = dst.deref_mut();
-            unwrap_registers!([self.constant_pool](src));
+            let mut dst = dst.entry_force_mut();
+            let dst = &mut **dst;
```
cedricjiang, line 29: may make sense for you to inline the line above into here?

this syntax is mind blowing :)

PR #33, opened 2025-04-28
Introduce a simple gas model that works

cedricjiang:


parasol_cpu/src/error.rs
```diff
@@ -197,6 +197,10 @@ pub enum Error {
     /// Attempted an operation that resulted in pointer overflow.
     #[error("32-bit Pointer overflow")]
     PointerOverflow,
+
+    /// Running out of allowed fee quota
+    #[error("Used gas amount {0} is exceeding quota {1}")]
+    OutOfGas(u32, u32),
```
cedricjiang, line 203: Not sure if it helps to add the instruction pointer where this happens...probably not?

parasol_cpu/src/proc/assembly.rs
```diff
@@ -173,81 +187,81 @@ define_op! {
     [Store (src dst, 1, PtrRegister) (src src, 0, Register) (meta width u32)],
 
     // And
-    [And (dst dst, 0, Register) (src a, 0, Register) (src b, 0, Register)],
+    [And (dst dst, 0, Register) (src a, 0, Register) (src b, 0, Register) (gas 1) (gas 100_000)],
```
cedricjiang, line 190: for "compute/compare" instructions only, two types of gas, one for "all plain" the other for "some cipher"

parasol_cpu/src/proc/assembly.rs
```diff
@@ -173,81 +187,81 @@ define_op! {
...
 
     // Or
-    [Or (dst dst, 0, Register) (src a, 0, Register) (src b, 0, Register)],
+    [Or (dst dst, 0, Register) (src a, 0, Register) (src b, 0, Register) (gas 1) (gas 100_000)],
 
     // Multiply a * b and produce the low word of the product.
-    [Mul (dst dst, 0, Register) (src a, 0, Register) (src b, 0, Register)],
+    [Mul (dst dst, 0, Register) (src a, 0, Register) (src b, 0, Register) (gas 1) (gas 500_000)],
```
cedricjiang, line 196: mul is "particularly expensive"

parasol_cpu/src/proc/mod.rs
```diff
@@ -548,6 +553,11 @@ impl FheProcessor {
 
         Ok(())
     }
+
+    fn get_gas_limit(&self) -> u32 {
+        // TODO: make this somehow configurable
+        100_000_000
```
cedricjiang, line 559: seems big enough to run my voting sample

parasol_cpu/src/tomasulo/tomasulo_processor.rs
```diff
@@ -151,7 +151,7 @@ macro_rules! impl_tomasulo {
 
                         let next_pc = self.next_program_counter(disp_inst, pc)?;
 
-                        Ok(next_pc)
+                        Ok((next_pc, *inst.get_gas_config().last().unwrap_or(&1)))
```
cedricjiang, line 154: TODO: this just assumes all operations are "some cipher", need to add code to figure out and choose the right one to use

parasol_cpu/src/proc/mod.rs
```diff
@@ -515,6 +516,90 @@ impl Tomasulo for FheProcessor {
             _ => Ok(pc + 1),
         }
     }
+
+    fn compute_gas(&self, dispatched_op: &crate::proc::DispatchIsaOp) -> u32 {
+        fn is_register_ciphertext(reg: &RobEntryRef<Register>) -> bool {
+            match reg {
+                RobEntryRef::Id(e) | RobEntryRef::IdMut(e) => e.entry().is_ciphertext(),
```
cedricjiang, line 523: assume this is what we want

parasol_cpu/src/proc/mod.rs
```diff
@@ -515,6 +516,90 @@ impl Tomasulo for FheProcessor {
...
+            }
+        }
+
+        use DispatchIsaOp::*;
+
+        match dispatched_op {
+            // instructions that do not compute anything are assigned trivial gas cost
+            BindReadOnly(..) | BindReadWrite(..) | Load(..) | LoadI(..) | Store(..)
+            | BranchNonZero(..) | BranchZero(..) | Cea(..) | Ceai(..) => 1,
```
cedricjiang, line 532: 1 for Cea ok?

parasol_cpu/src/proc/mod.rs
```diff
@@ -515,6 +516,90 @@ impl Tomasulo for FheProcessor {
...
+
+            // instructions that compute on one input source, but gas does not rely on it
+            Zext(..) | Trunc(..) => 1,
```
cedricjiang, line 535: the actual width is plain text so it's kind of "fast" (use the `convert` method), but is 1 good?

parasol_cpu/src/proc/mod.rs
```diff
@@ -515,6 +516,90 @@ impl Tomasulo for FheProcessor {
...
+            | CmpLt(_, input1, input2)
+            | CmpLe(_, input1, input2) => {
+                if is_register_ciphertext(input1) || is_register_ciphertext(input2) {
+                    100_000
+                } else {
+                    1
+                }
+            }
+
+            Mul(_, input1, input2) => {
```
cedricjiang, line 564: special, more expensive for now

parasol_cpu/src/proc/mod.rs
```diff
@@ -515,6 +516,90 @@ impl Tomasulo for FheProcessor {
...
+                if is_register_ciphertext(input1) || is_register_ciphertext(input2) {
+                    500_000
+                } else {
+                    1
+                }
+            }
+
+            // instructions that compute on two input sources that are not interchangeable, and gas relies on only one of them
+            Shr(_, _, input)
```
cedricjiang, line 573: shift/rotation consumption of CPU resources only relies on `input` (here the shift amount) being ciphertext or not

parasol_cpu/src/proc/mod.rs
```diff
@@ -515,6 +516,90 @@ impl Tomasulo for FheProcessor {
...
+                    || is_register_ciphertext(input3)
+                {
+                    100_000
+                } else {
+                    1
+                }
+            }
+
+            // return has zero gas cost
+            Ret() => 0,
```
cedricjiang, line 600: 0 makes more sense than 1 (if 1, think that you are right at the threshold, and executing `Ret` gets you out of gas - not good)

parasol_cpu/src/tomasulo/registers.rs
```diff
@@ -84,7 +84,7 @@ pub struct IdRobEntry<T>
 where
     T: Default,
 {
-    register: Arc<RwLock<RobEntry<T>>>,
+    pub(crate) register: Arc<RwLock<RobEntry<T>>>,
```
cedricjiang, line 87: needed for inspecting the type

parasol_cpu/src/tomasulo/tomasulo_processor.rs
cedricjiang: done...

parasol_cpu/src/proc/mod.rs
```diff
@@ -548,6 +638,11 @@ impl FheProcessor {
 
         Ok(())
     }
+
+    fn get_gas_limit(&self) -> u32 {
+        // TODO: make this somehow configurable
+        100_000_000
```
cedricjiang, line 644: my voting sample uses 12_200_141

parasol_cpu/src/error.rs
rickwebiii: You should see the error messages for actual block chains... This is probably more than sufficient.

parasol_cpu/src/proc/mod.rs
```diff
@@ -515,6 +516,90 @@ impl Tomasulo for FheProcessor {
...
+                    100_000
+                } else {
+                    1
+                }
+            }
+
+            // instructions that compute on three input sources that are interchangeable, and gas relies on either of them
+            AddC(_, _, input1, input2, input3)
+            | SubB(_, _, input1, input2, input3)
+            | Cmux(_, input1, input2, input3) => {
```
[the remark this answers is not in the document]
rickwebiii: It's fine for now.

parasol_cpu/src/proc/mod.rs
rickwebiii: Address the TODO and take the gas limit in the run function. Then write some tests that use too much gas.

rickwebiii (approved):

parasol_cpu/src/proc/mod.rs
rickwebiii: This instruction is getting deleted.

cedricjiang: ok cool, so here only load/store/branch will stay?

parasol_cpu/benches/programs/chi_squared.rs
```diff
@@ -251,6 +251,7 @@ where
             output_buffer_2,
             output_buffer_3,
         ],
+        5_000_000,
```
cedricjiang, line 254: this method seems unused...but put in 5M anyway, same as the other

parasol_cpu/src/proc/mod.rs
```diff
@@ -998,4 +1102,31 @@ mod buffer_uint_tests {
...
+                IsaOp::BindReadWrite(RegisterName::new(1), 1, true),
+                IsaOp::Load(RegisterName::new(0), RegisterName::new(0), 8),
+                IsaOp::Not(RegisterName::new(1), RegisterName::new(0)),
+                IsaOp::Store(RegisterName::new(1), RegisterName::new(1), 8),
+            ],
+        };
+
+        assert_eq!(
+            cpu.run_program(not_program, &buffers, 100).err().unwrap(),
+            Error::OutOfGas(100_003, 100)
```
cedricjiang, line 1129: too precise :)

cedricjiang: this is hit at NOT which is 100k plus three bind/load above

PR #34, opened 2025-04-29
Unified memory model

rickwebiii:


parasol_cpu/benches/programs/chi_squared.rs
```diff
@@ -1,349 +1,335 @@
-use parasol_cpu::{
```
rickwebiii, line 1: Should fix this benchmark

parasol_cpu/src/proc/assembly.rs
```diff
@@ -8,10 +8,11 @@ macro_rules! define_op {
         $dispatch_name:ident,
         ($($reg_kind:ty),*),
```
rickwebiii, line 9: Include the register count here somehow so we can derive the mask and shifts below

parasol_cpu/src/proc/assembly.rs
```diff
@@ -43,9 +44,81 @@ macro_rules! define_op {
...
+                        match x {
+                            $(
+                                $inst_name::$op_name (
+                                    $($dst_name,)*
+                                    $($src_name,)*
+                                    $($meta_name,)*
+                                ) => {
+                                    $(
+                                        encoded |= ($dst_name.name as u64) << shift;
+                                        shift += 5;
```
rickwebiii, line 102: Parasol today has 64 registers in the encoded instruction, not 32. So 6, not 5. Ideally, we pass the count and derive this value.

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -0,0 +1,707 @@
...
+                    }
+                } else {
+                    Err(Error::BranchConditionNotPlaintext)
+                }
+            }
+            DispatchIsaOp::Ret() => Err(Error::Halt),
+            _ => Ok(pc + 8),
+        }
+    }
+}
```
rickwebiii, line 707: Add tests to verify the call ABI

parasol_cpu/src/proc/ops/bitshift.rs
```diff
@@ -105,7 +105,7 @@ impl FheProcessor {
         src: RobEntryRef<Register>,
         shift: RobEntryRef<Register>,
         _instruction_id: usize,
-        _pc: usize,
+        _pc: u32,
```
rickwebiii, line 108: Remove

parasol_cpu/src/error.rs
```diff
@@ -197,6 +198,30 @@ pub enum Error {
     /// Attempted an operation that resulted in pointer overflow.
     #[error("32-bit Pointer overflow")]
     PointerOverflow,
+
+    /// Invoked an FHE program with more than 8 arguments.
+    #[error(
+        "FHE Programs may have up to 8 arguments. This limitation will be removed in the future."
+    )]
+    NotImplementedTooManyArgs,
```
ryanorendorff, line 206: What causes this?

parasol_cpu/src/proc/assembly.rs
```diff
@@ -155,100 +291,91 @@ macro_rules! define_op {
...
-    [BindReadWrite (dst dst, 1, PtrRegister) (meta id usize) (meta encrypted bool)],
+    (Register),
 
     // Load
-    [Load (dst dst, 0, Register) (src src, 1, PtrRegister) (meta width u32)],
+    [0x02 Load (dst dst, 0, Register) (src src, 0, Register) (meta width, 7, u32)],
 
     // Load immediate
-    [LoadI (dst dst, 0, Register) (meta imm u128) (meta width u32)],
+    [0x03 LoadI (dst dst, 0, Register) (meta imm, 44, u128) (meta width, 7, u32)],
```
ryanorendorff, line 300: How will large immediates get handled later?

parasol_cpu/src/proc/assembly.rs
```diff
@@ -155,100 +291,91 @@ macro_rules! define_op {
...
 
     // Store
-    [Store (src dst, 1, PtrRegister) (src src, 0, Register) (meta width u32)],
+    [0x04 Store (src dst, 0, Register) (src src, 0, Register) (meta width, 7, u32)],
```
ryanorendorff, line 303: I assume the opcode numbers come from the compiler; potentially add a note on where to change them in that codebase.

parasol_cpu/src/proc/ops/load.rs
```diff
@@ -1,91 +1,98 @@
...
-
-                    let range = ptr.offset as usize * 8..ptr.offset as usize * 8 + width as usize;
-
-                    for (input, output) in ptr.base[range].iter().zip(result.iter()) {
-                        let input = AtomicRefCell::borrow(input);
-                        let mut output = AtomicRefCell::borrow_mut(output);
-
-                        output.clone_from(&input);
-                    }
+                    // Plaintext loads and stores exec immediately, so we don't need to update our deps.
```
ryanorendorff, line 38: Is the word "Plaintext" valid in this comment? The ciphertext case also seems to execute immediately.

parasol_cpu/src/proc/ops/store.rs
```diff
@@ -1,117 +1,72 @@
...
-
-                    if val.len() != width as usize {
-                        return Err(Error::WidthMismatch {
-                            inst_id: instruction_id,
-                            pc,
-                        });
+                    if base_addr % num_bytes != 0
+                        || num_bytes > 16
+                        || !num_bytes.is_power_of_two()
+                        || num_bytes == 0
```
ryanorendorff, line 34: Since these conditions are reused between load and store it might be good to abstract them into one function in case we need to change the conditions.

ryanorendorff (approved):

examples/basic_add/src/main.rs
```diff
@@ -1,46 +1,48 @@
...
+
+//     // To pass arguments into the Parasol C program, we must convert
+//     // them to `Buffer`s. Note that we must provide an output
+//     // buffer as well!
+//     let arguments =
+//         [a, b, 0u8].map(|x| Buffer::cipher_from_value(&x, &Encryption::default(), &secret_key));
+
+//     // Run the program.
+//     let now = Instant::now();
+//     let encrypted_result = run_program(compute_key.clone(), FHE_FILE, "add", &arguments).unwrap();
```
cedricjiang, line 34: hint: merge conflict, because of gas limit added

parasol_cpu/benches/chi_squared.rs
```diff
@@ -1,72 +1,73 @@
...
+//     let n_2 = 3u16;
+//     let zero = 0u16;
 
-    let optimised_program = chi_squared_optimised_program(bit_width, encrypted_computation);
+//     let optimised_program = chi_squared_optimised_program(bit_width, encrypted_computation);
 
-    let buffer_0 = buffer_from_value_80(n_0, &enc, encrypted_computation);
-    let buffer_1 = buffer_from_value_80(n_1, &enc, encrypted_computation);
-    let buffer_2 = buffer_from_value_80(n_2, &enc, encrypted_computation);
+//     // let buffer_0 = buffer_from_value_80(n_0, &enc, encrypted_computation);
```
cedricjiang, line 23: two levels of comments 🤣

parasol_cpu/benches/chi_squared.rs
```diff
@@ -1,72 +1,73 @@
...
-    //             ],
-    //         )
-    //         .unwrap();
-    //     });
-    // });
-}
+//     // Seems to stall sometimes
+//     // group.bench_function("chi squared naive", |b| {
+//     //     b.iter(|| {
+//     //         proc.run_program(
```
cedricjiang, line 56: merge conflict with gas limit

parasol_cpu/benches/programs/chi_squared.rs
```diff
@@ -1,349 +1,335 @@
...
+//         buffer_0.clone(),
+//         buffer_1.clone(),
+//         buffer_2.clone(),
+//         output_buffer_0.clone(),
+//         output_buffer_1.clone(),
+//         output_buffer_2.clone(),
+//         output_buffer_3.clone(),
+//     ];
+
+//     proc.run_program(
```
cedricjiang, line 229: merge conflict with gas limit

parasol_cpu/src/error.rs
```diff
@@ -3,17 +3,21 @@ use std::ffi::FromBytesUntilNulError;
 use elf::ParseError;
 use thiserror::Error;
 
-#[derive(Debug, Clone, Error, PartialEq, Eq)]
+#[derive(Debug, Clone, Error)]
```
cedricjiang, line 6: any reason for removing `PartialEq`? some assertions rely on this, like in gas limit test

cedricjiang: speaking of that, the new `OutOfGas` error type will be a merge conflict too

parasol_cpu/src/memory/mod.rs
```diff
@@ -29,19 +29,151 @@ const TOTAL_PAGES: u32 = 0x1 << (32 - LOG2_PAGE_SIZE);
...
+                    .collect::<Vec<_>>();
+                Byte::try_from(zero).unwrap()
+            }
+            (Byte::Plaintext(val), Extend::Signed) => {
+                let sext = if val >> 7 == 0x1 { 0xFFu8 } else { 0 };
+                Byte::from(sext)
+            }
+            (Byte::Ciphertext(val), Extend::Signed) => {
+                let sext = (0..8)
+                    .map(|_| val.last().unwrap().clone())
```
cedricjiang, line 126: last or first

cedricjiang: (just asking, I assume it's right i.e. matches endianness)

parasol_cpu/src/memory/mod.rs
```diff
@@ -121,6 +278,70 @@ impl Memory {
...
+            stack_ptr: Mutex::new(stack_start.try_offset(stack_size).unwrap()),
+        };
+
+        memory.try_allocate_at(stack_start, stack_size).unwrap();
+
+        memory
+    }
+
+    pub(crate) fn new_default_stack() -> Self {
+        Memory::new(Ptr32::from(0x8000_0000), 4096)
```
cedricjiang, line 295: this size is...fairly small

cedricjiang: assume it's just "initial"

parasol_cpu/src/memory/mod.rs
```diff
@@ -208,14 +470,167 @@ impl Memory {
...
+        for i in start_page..=end_page {
+            if pages[i as usize].is_none() {
+                return Err(Error::AccessViolation(ptr.try_offset(i * PAGE_SIZE)?.0));
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Attempt to write `x` to the pre-allocated address given by `ptr`.
```
cedricjiang, line 534: this is the same text as `try_write_type` above

parasol_cpu/src/memory/mod.rs
```diff
@@ -208,14 +470,167 @@ impl Memory {
...
+    ///
+    /// # Remarks
+    /// An error occurs if allocation fails.
```
cedricjiang, line 537: "Returns an error if..."

parasol_cpu/src/memory/mod.rs
```diff
@@ -208,14 +470,167 @@ impl Memory {
...
+        self.try_write_type(ptr, x)?;
+
+        Ok(ptr)
+    }
 }
 
+/// An 8-bit encrypted or plaintext value.
 #[derive(Clone)]
 pub enum Byte {
+    /// An un-encrypted 8-bit value.
```
cedricjiang, line 550: maybe just say plaintext...

parasol_cpu/src/memory/mod.rs
```diff
@@ -208,14 +470,167 @@ impl Memory {
...
+            Self::Plaintext(val) => *val,
+            _ => panic!("Not a plaintext"),
+        }
+    }
+
+    /// Unwrap the inner ciphertext data or panic if it's a plaintext.
+    ///
+    /// # Panics
+    /// If the byte contains plaintext data.
+    pub fn unwrap_ciphertext(&self) -> Vec<Arc<AtomicRefCell<L1GlweCiphertext>>> {
```
cedricjiang, line 601: I think the unwrap convention is to consume the value itself...

cedricjiang: more relevant here than plaintext

parasol_cpu/src/error.rs
rickwebiii: Limitation was removed :)

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+    /// The number of bytes this type takes. Should be no greater than `u32::MAX`.
+    const SIZE: usize;
+
+    /// Whether this type needs to be sign extended or not.
+    const SIGNED: bool;
+
+    /// Convert a byte array and metadata into a [`Vec<Arg>`].
+    ///
+    /// # Remarks
+    /// This positions arguments
```
cedricjiang, line 27: what does this mean?

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+    /// Convert this value into a [`Vec<Byte>`].
+    ///
+    /// # Remarks
+    /// The number of bytes returned must equal `T::size` or panics may result in related
+    /// methods.
+    fn to_bytes(&self) -> Vec<Byte>;
+
+    /// Convert this value into a [`Vec<Arg>`]` for calling with a function.
```
cedricjiang, line 66: extra backtick after right bracket

parasol_cpu/src/proc/assembly.rs
rickwebiii: There are a few ways one does this on real ISAs:
* Risc-V allows 16-bit immediates. You can use li to load into the lower 16-bits of a 32-bit word and liu to load into the upper 16 bits of a word. So, it takes 2 instructions to load a 32-bit immediate.
* You can load values from the .rodata segment (which we now have)

I'm inclined to just require large constants get loaded from constant memory.

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+    ///
+    /// # Panics
+    /// If `self.to_bytes().len() != Self::SIZE`.
+    fn to_args(&self) -> Vec<Arg> {
+        let bytes = self.to_bytes();
+
+        assert_eq!(
+            bytes.len(),
+            Self::SIZE,
+            "Byte length mismatch, expected {} got {}",
```
cedricjiang, line 80: not sure if you need to repeat the two args...assert will report left and right, so you _could_ just say "expected left got right"

just saying

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+}
+
+macro_rules! primitive_impl_to_arg {
+    ($t:ty,$sext:literal) => {
+        paste! {
+            impl ToArg for $t {
+                const ALIGNMENT: usize = std::mem::align_of::<$t>();
+                const SIZE: usize = std::mem::size_of::<$t>();
+                const SIGNED: bool = $sext;
+
```
cedricjiang, line 113: extra empty line

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Attempt to create a value of this type from the given bytes.
+    fn try_from_bytes(data: Vec<Byte>) -> Result<Self>;
+}
+
+macro_rules! primitive_impl_to_arg {
+    ($t:ty,$sext:literal) => {
```
cedricjiang, line 107: sext should be "signedness", just saying

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+primitive_impl_to_arg!(u32, false);
+primitive_impl_to_arg!(u64, false);
+primitive_impl_to_arg!(u128, false);
+primitive_impl_to_arg!(i8, true);
+primitive_impl_to_arg!(i16, true);
+primitive_impl_to_arg!(i32, true);
+primitive_impl_to_arg!(i64, true);
+primitive_impl_to_arg!(i128, true);
+
+impl<const N: usize> ToArg for UInt<N, L1GlweCiphertext> {
```
cedricjiang, line 181: no signed version?

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+            return Err(Error::TypeSizeMismatch);
+        }
+
+        let data = data
+            .into_iter()
+            .map(|x| match x {
+                Byte::Plaintext(_) => Err(Error::EncryptionMismatch),
+                Byte::Ciphertext(val) => Ok(val),
+            })
+            .collect::<Result<Vec<_>>>()?
```
cedricjiang, line 206: do you have to collect here?

parasol_cpu/src/error.rs
```diff
@@ -3,17 +3,21 @@ use std::ffi::FromBytesUntilNulError;
...
+
     /// Illegal operands to an instruction.
     #[error("(inst:{inst_id}, pc:0x{pc:x}) Illegal operands executing instruction at PC")]
     IllegalOperands {
         /// The faulting instruction's id.
         inst_id: usize,
 
         /// The program counter at time of error.
-        pc: usize,
+        pc: u32,
```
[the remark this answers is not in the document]
rickwebiii: This was a journey into lands unknown. I'd rather not break all the refactorings out.

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+        if T::SIZE == 0 {
+            return vec![];
+        }
+
+        self.iter()
+            .flat_map(|x| {
+                // Pad each array element to its alignment...
+                let bytes = x.to_bytes();
+
+                bytes
```
cedricjiang, line 230: any reason this variable is not inlined

parasol_cpu/src/memory/mod.rs
rickwebiii: Last byte contains the sign bit, which we need to extend.

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+
+/// A type for passing arguments to Parasol programs. When invoking
+/// [`crate::FheComputer::run_program`], the processor will transparently set up the registers
+/// and first stack frame according to Parasol ABI's calling convention.
+///
+/// # Remarks
+/// Parasol follows the ILP32 RISC-V [calling convention](https://drive.google.com/file/d/1Ja_Tpp_5Me583CGVD-BIZMlgGBnlKU4R/view?pli=1).
+///
+/// 4-byte or smaller scalars go in the next available a0-a7 register. Smaller signed/unsigned
+///  scalars are sign/zero extended to 32-bit (respectively). If no registers remain, they get
```
cedricjiang, line 295: extra whitespace

parasol_cpu/src/memory/mod.rs
rickwebiii: This method only exists for testing so we can run programs without ELF files. We'll parse actual stack sizes from ELF binaries.

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+/// 5-8 byte scalars go in the next 2 available registers. The lo word
+/// goes into the first and the high word into the second. Signed/unsigned scalars are
+/// sign/zero extended to 32-bits (respectively). If one register remains, the lo word goes into the
+/// register while the high word goes on the stack. If no registers remain, they both get pushed onto
+/// the stack, where they will be aligned but not extended.
+///
+/// 1-4 byte aggregates get packed into a single register (if available) or placed on the stack
+/// according to its alignment requirements.
+///
+/// 5-8 byte aggregate get packed into 2 registers (if available). If only one register is available,
```
cedricjiang, line 307: aggregates

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+/// available, both words get pushed on the stack.
+///
+/// Scalars or aggregates larger than 64-bits will be transparently allocated on the
+/// Parasol heap and passed by reference (i.e. a 32-bit pointer will be passed for the corresponding
+/// register/stack argument).
+pub struct ArgsBuilder {
+    args: Vec<Arg>,
+}
+
+impl Default for ArgsBuilder {
```
cedricjiang, line 318: I wonder if this can be just #[derive(Default)]

cedricjiang: same effect I think

parasol_cpu/src/proc/args.rs
```diff
@@ -0,0 +1,458 @@
...
+
+        self
+    }
+
+    /// Specify a return value for an FHE program.
+    ///
+    /// # Remarks
+    /// If an FHE program returns a value greater than 8 bytes, you must specify this, even if you
+    /// don't use it.
+    /// Failure to do so will result in incorrect execution.
```
cedricjiang, line 342: bummer...

parasol_cpu/src/proc/assembly.rs
```diff
@@ -43,9 +44,81 @@ macro_rules! define_op {
...
+                        Ok(match [<$inst_name OpCode>]::try_from((value & 0xFF) as u8)? {
+                            $(
+                                [<$inst_name OpCode>]::$op_name => {
+                                    #[allow(unused)]
+                                    let value = value >> 8;
+
+                                    $(
+                                        let $dst_name = RegisterName::new(value as usize & 0x1F);
+                                        #[allow(unused)]
+                                        let value = value >> 5;
```
cedricjiang, line 61: I think it's 6, unless we want to change from both compiler and `parse_elf`

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -0,0 +1,776 @@
...
+    atomic::Ordering,
+    mpsc::{self, Receiver, Sender},
+};
+
+pub(crate) struct FheProcessor
+where
+    Self: Tomasulo,
+{
+    /// The register file.
+    registers: RegisterFile<Register, DispatchIsaOp>,
```
cedricjiang, line 25: A little surprised that register is "associated" with dispatched instruction...

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -0,0 +1,776 @@
...
+        register_config: &RegisterConfig,
+        aux_data: <Self as Tomasulo>::AuxiliaryData,
+    ) -> Self {
+        let registers = RegisterFile::<Register, DispatchIsaOp>::new(register_config.num_registers);
+
+        Self {
+            registers,
+            aux_data,
+            pc: 0,
+            current_instruction: 0usize,
```
cedricjiang, line 67: I don't think we need the suffixes? given the type is very well specified...

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -0,0 +1,776 @@
...
+            instructions_inflight: 0usize,
+            ready_instructions: mpsc::channel(),
+        }
+    }
+
+    // This method requires the assumption that load and store operations have no in-flight
```
cedricjiang, line 73: "require the assumption" reads a bit unusual

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -0,0 +1,776 @@
...
+
+                update_memory_deps(src, *num_bytes)?
+            }
+            _ => {}
+        };
+
+        Ok(())
+    }
+
+    pub fn dispatch_instruction(&mut self, inst: IsaOp, pc: u32) -> Result<u32> {
```
cedricjiang, line 146: are you going to move the gas stuff here...

parasol_cpu/src/proc/tests/add.rs
```diff
@@ -11,51 +13,29 @@ use parasol_runtime::test_utils::get_secret_keys_80;
...
             IsaOp::Add(
-                RegisterName::new(2),
-                RegisterName::new(0),
-                RegisterName::new(1),
+                RegisterName::new(10),
+                RegisterName::new(10),
+                RegisterName::new(11),
             ),
-            IsaOp::Store(RegisterName::new(2), RegisterName::new(2), 32),
+            IsaOp::Ret(),
```
cedricjiang, line 28: do we now require an explicit return?

parasol_cpu/src/proc/tests/load_store.rs
```diff
@@ -1,218 +1,174 @@
...
-        {
-            for (j, out) in out.iter().enumerate() {
-                let expected = (plain >> j) & 0x1;
-                let actual = enc.decrypt_glwe_l1(out, &get_secret_keys_80()).coeffs()[0];
-                assert_eq!(expected as u64, actual);
-            }
+            .map(|x| UInt::<8, L1GlweCiphertext>::encrypt_secret(*x as u64, &enc, &sk))
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap_or_else(|_| unreachable!());
```
cedricjiang, line 76: could just `unwrap`? unless you want `unreachable!` other than `panic!`

parasol_cpu/src/error.rs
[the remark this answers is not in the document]
cedricjiang: ok I realize the change is because now "pc" is actually "location", not just "counter" (I see it increases by instruction length 8 instead of just 1) that makes sense

parasol_cpu/src/error.rs
cedricjiang: I realize the test can change to use `matches!` macro, anyway

parasol_cpu/src/error.rs
```diff
@@ -101,7 +99,7 @@ pub enum Error {
...
```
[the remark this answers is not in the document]
rickwebiii: pc is an actual pointer now. We load instructions from the unified memory.

parasol_cpu/src/proc/args.rs
rickwebiii: We'll add it in another review...

parasol_cpu/src/proc/args.rs
rickwebiii: It's a new paragraph.

parasol_cpu/src/proc/args.rs
rickwebiii: We don't even need this trait. Dunno why I did this.

parasol_cpu/src/proc/args.rs
rickwebiii: This is an artifact of ILP32's calling convention. The caller must provide space (usually on the stack, but we use the heap) for > 8-byte return values and pass a pointer to the allocated space in x10. Thus, the callee expects this to be non-null and correctly setting the return value accomplishes this.

Perhaps we can solve this later if we can query the call signature.

parasol_cpu/src/proc/assembly.rs
rickwebiii: Will fix in a follow up review.

parasol_cpu/src/proc/fhe_processor.rs
rickwebiii: We need to look up the last instruction that wrote to this named register so we can add a dependency. I guess we could store this in a separate structure, but this is what we did.

parasol_cpu/src/proc/fhe_processor.rs
rickwebiii: Done

parasol_cpu/src/proc/tests/add.rs
rickwebiii: It was previously not required because we didn't use emulated memory for storing the program and would just return automatically if you jumped outside the program buffer.

Now, we're actually fetching instructions from unified memory and you'll get a decode failure if you run beyond the end of a program. So, it's absolutely now needed.

parasol_cpu/src/proc/tests/load_store.rs
rickwebiii: Actually, we can't because the thing we're unwrapping doesn't impl `Debug`, which unwrap requires.

parasol_cpu/src/proc/assembly.rs
rickwebiii: Another review.

parasol_cpu/src/proc/args.rs
cedricjiang: not quite sure what you mean, but I mean
```
///  scalars
```
should be
```
/// scalars
```
I don't think it's a new paragraph though...

parasol_cpu/src/proc/args.rs
cedricjiang: ok...

parasol_cpu/src/proc/args.rs
cedricjiang: that's fine, but if there's some "internal" way to solve this without asking the user to do this it'll be nicer

parasol_cpu/src/proc/tests/add.rs
cedricjiang: ok, I see many tests did not have explicit return, something to deal with later I think...

parasol_cpu/src/proc/tests/load_store.rs
cedricjiang: I see...

PR #42, opened 2025-05-02
Add signed extension support to the processor

cedricjiang:


parasol_cpu/src/proc/assembly.rs
```diff
@@ -368,6 +368,7 @@ define_op! {
     // Casting operations
     [0x05 Zext (dst dst, 0, Register) (src src, 0, Register) (meta width, 7, u32)],
     [0x06 Trunc (dst dst, 0, Register) (src src, 0, Register) (meta width, 7, u32)],
+    [0x07 Sext (dst dst, 0, Register) (src src, 0, Register) (meta width, 7, u32)],
```
cedricjiang, line 371: I'd like to do 0x04 but it's not available

parasol_cpu/src/proc/ops/casting.rs
```diff
@@ -21,7 +21,48 @@ impl FheProcessor {
...
+
+    pub fn ext(
+        &mut self,
+        retirement_info: RetirementInfo<DispatchIsaOp>,
+        dst: RobEntryRef<Register>,
+        src: RobEntryRef<Register>,
+        new_width: u32,
+        instruction_id: usize,
+        pc: u32,
+        signed: bool,
```
cedricjiang, line 63: code shared between signed and unsigned

parasol_cpu/src/proc/ops/casting.rs
```diff
@@ -32,25 +73,38 @@ impl FheProcessor {
             }
 
             match src {
-                Register::Plaintext { val, .. } => {
+                Register::Plaintext { val, width } => {
                     *dst = Register::Plaintext {
-                        val: *val,
+                        val: if signed {
+                            let sign_mask = 1 << (*width - 1);
```
cedricjiang, line 79: for the given "old" width, compute the sign mask, for example if width is 4, sign mask should be 0b1000

parasol_cpu/src/proc/ops/casting.rs
```diff
@@ -32,25 +73,38 @@ impl FheProcessor {
...
+                            if (sign_mask & *val) == 0 {
+                                *val
```
cedricjiang, line 81: positive, as is

parasol_cpu/src/proc/ops/casting.rs
```diff
@@ -32,25 +73,38 @@ impl FheProcessor {
...
+                            } else {
+                                let abs = (sign_mask << 1) - val;
```
cedricjiang, line 83: `sign_mask << 1` is essentially `1 << *width`, so this turn a negative number into its absolute value, so later it can be subtracted from a bigger 0b10000....000 next line

parasol_cpu/src/proc/ops/casting.rs
```diff
@@ -32,25 +73,38 @@ impl FheProcessor {
...
-
                     FheProcessor::retire(&retirement_info, Ok(()));
                 }
                 Register::Ciphertext(Ciphertext::L1Glwe { data }) => {
                     let current_width = data.len() as u32;
 
                     // Get trivial zeros
-                    let zero = Arc::new(AtomicRefCell::new(self.aux_data.l1glwe_zero.clone()));
+                    let pad = if signed {
+                        data.last().unwrap().clone()
```
cedricjiang, line 98: last is sign bit, clone should work here, and I assume it's ok for later `repeat_n`?

parasol_cpu/src/proc/assembly.rs
ryanorendorff: Just renumber things to make sense. There will be breaking ABI changes anyway.

parasol_cpu/src/proc/ops/casting.rs
ryanorendorff: Yes that should be fine.

parasol_cpu/src/proc/ops/casting.rs
ryanorendorff: I believe you can do this as well, which makes the result a little easier to see that the value is being padded with 1s in the case of a signed number.
```rust
    let mask = (1u128.checked_shl(width).unwrap_or(0)).wrapping_sub(1);
    let new_mask = (1u128.checked_shl(new_width).unwrap_or(0)).wrapping_sub(1);
    let sign_bit = (value >> (width - 1)) & 0x1;
    
    // We should probably have a contention on what is in the unused bits but masking makes that irrelevant.
    let value = value & mask; 
    
    if value & sign_bit != 0 {
        // pad with the sign bit
        (value | (!mask)) & new_mask
    } else {
        value
    }
```

ryanorendorff (changes requested):
In general looks like the right path. Needs tests for the all cases (zext and sext).

parasol_cpu/src/proc/assembly.rs
cedricjiang: Technically it'll could be compatible though, I will just pick "random number" for now and we can make a batch change when the ISA stablizes...

parasol_cpu/src/proc/ops/casting.rs
cedricjiang: wow that's so rusty...

cedricjiang:
I need to find where the tests are :)

parasol_cpu/src/proc/ops/casting.rs
cedricjiang: so `sign_bit` is actually either `1` or `0` right, not `1` or `0` on the sign position. that is if "old" width is 4 and value is 0b1001, then `sign_bit` is actually `1` not `0b1000`, this means the if check actually checks for even/odd not positive/negative...

cedricjiang: actually you should just check if `sign_bit` itself is 0 or not, its "the sign bit of the value", not some "mask"

cedricjiang: I do like your pad approach though :)

parasol_cpu/src/proc/ops/casting.rs
```diff
@@ -77,12 +77,12 @@ impl FheProcessor {
...
-                            let sign_mask = 1 << (*width - 1);
-                            if (sign_mask & *val) == 0 {
+                            if val >> (width - 1) == 0 {
                                 *val
                             } else {
-                                let abs = (sign_mask << 1) - val;
-                                (1 << new_width) - abs
+                                let old_mask = (1 << width) - 1;
+                                let new_mask = (1 << new_width) - 1;
+                                (val | !old_mask) & new_mask
```
cedricjiang, line 85: @ryanorendorff took your approach but assume valid input so do not need the verbosity

parasol_cpu/src/proc/tests/casting.rs
```diff
@@ -12,19 +12,26 @@ use crate::{
...
         .iter()
         .cartesian_product(supported_sizes.iter())
         .map(|(input_width, output_width)| {
             (
-                (*input_width, *output_width),
-                if zero_extend {
-                    input_width <= output_width
-                } else {
-                    input_width >= output_width
+                (*input_width, *output_width as u32),
```
cedricjiang, line 29: this cast is needed for `saturating_sub` to work later

parasol_cpu/src/proc/tests/casting.rs
```diff
@@ -38,15 +45,21 @@ fn casting(zero_extend: bool, encrypted_computation: bool) {
...
         let enc = &enc;
 
-        // Get a random 32 bit value
-        let value = thread_rng().next_u32();
-        let shift_width = if zero_extend {
-            input_width
-        } else {
-            output_width
+        // Get a random value with input width
+        let value = thread_rng().next_u32() & ((1 << input_width) - 1);
```
cedricjiang, line 49: I don't get why `value` wasn't truncated to the input width in the first place...so we allow value = 0b10101010 and width = 4 to mean the same thing as value = 0b1010 and width = 4? that sounds very error-prone to me

parasol_cpu/src/proc/tests/casting.rs
```diff
@@ -38,15 +45,21 @@ fn casting(zero_extend: bool, encrypted_computation: bool) {
...
+        let expected = match cast_type {
+            CastType::SignExtension => {
+                let mut sign_bit = value & (1 << (input_width - 1));
+                let mut extended = value;
+                for _ in 0..output_width.saturating_sub(input_width) {
+                    sign_bit <<= 1;
+                    extended |= sign_bit;
+                }
+                extended
```
cedricjiang, line 58: intentionally a different sign extension method (this is a literal one)

parasol_cpu/src/proc/tests/casting.rs
cedricjiang: damn...linter does not like this...

parasol_cpu/src/proc/tests/casting.rs
```diff
@@ -12,19 +12,26 @@ use crate::{
...
-fn casting(zero_extend: bool, encrypted_computation: bool) {
-    let supported_sizes = [8, 16, 32];
+enum CastType {
+    ZeroExtension,
+    SignExtension,
+    Truncation,
+}
+
+fn casting(cast_type: CastType, encrypted_computation: bool) {
+    let supported_sizes = [8u32, 16, 32];
```
cedricjiang, line 22: u32 suffix needed now to make `saturating_sub` and linter later happy...

parasol_cpu/src/proc/assembly.rs
ryanorendorff: Ya that is fine, I imagine we will change things up when the memory changes land.

parasol_cpu/src/proc/ops/casting.rs
ryanorendorff: Sounds good.

parasol_cpu/src/proc/tests/casting.rs
ryanorendorff: It is likely better to be defensive on the inputs than make an assumption.

ryanorendorff (approved):

parasol_cpu/src/proc/tests/casting.rs
cedricjiang: I vaguely remember there is code that treats this as illegal...I may get this back for sign extension but will create an issue for that

PR #44, opened 2025-05-05
Add signed integer test support

cedricjiang:


parasol_cpu/src/proc/args.rs
```diff
@@ -175,6 +178,40 @@ impl<const N: usize> ToArg for UInt<N, L1GlweCiphertext> {
     }
 }
 
+impl<const N: usize> ToArg for Int<N, L1GlweCiphertext> {
```
cedricjiang, line 181: this is the impl copied from `UInt`, only change seems to be `SIGNED` needs to be true, let me know if I am missing anything here. this is not just test code

parasol_cpu/src/test_utils.rs
```diff
@@ -59,13 +59,48 @@ pub fn make_computer_128() -> (FheComputer, Encryption) {
...
+    fn test_from(value: u64) -> Self {
+        value as i16
+    }
+}
+
+impl TestFrom<u64> for i8 {
+    fn test_from(value: u64) -> Self {
+        value as i8
+    }
+}
```
cedricjiang, line 100: the entire "TestFrom" stuff is very nasty and I wonder if there's a cleaner way...

cedricjiang: the reason for that is we do want `u64` to be used for signed, here `u64` has no signedness meaning - it's just used as "raw bytes", think `[bool; 64]`

parasol_cpu/src/test_utils.rs
```diff
@@ -59,13 +59,48 @@ pub fn make_computer_128() -> (FheComputer, Encryption) {
...
+
 pub trait Bits<const N: usize> {
-    type PlaintextType: num::Num
-        + TryFrom<u64, Error = std::num::TryFromIntError>
-        + Into<u64>
-        + std::fmt::Debug
-        + Copy
-        + ToArg;
+    type PlaintextType: num::Num + TestFrom<u64> + std::fmt::Debug + Copy + ToArg;
```
cedricjiang, line 103: move `Into<u64>` out as signed version will do `i64`

parasol_cpu/src/test_utils.rs
```diff
@@ -102,10 +139,11 @@ where
 impl<const N: usize> MaybeEncryptedUInt<N>
 where
     BitsUnsigned: Bits<N>,
+    <BitsUnsigned as Bits<N>>::PlaintextType: Into<u64>,
 {
     pub fn new(val: u64, enc: &Encryption, sk: &SecretKey, encrypt: bool) -> Self {
         if !encrypt {
-            Self::Plain(val.try_into().unwrap())
+            Self::Plain(<BitsUnsigned as Bits<N>>::PlaintextType::test_from(val))
```
cedricjiang, line 146: a bit verbose, could do a `TestInto` sugar but whatever

parasol_cpu/src/test_utils.rs
```diff
@@ -150,11 +189,93 @@ where
     }
 }
 
+pub struct BitsSigned();
```
cedricjiang, line 192: the signed double from here, changes include:
1) inner plaintext type are signed
2) enum encrypted variant has int instead of uint
3) ToArg impl has true SIGNED
hopefully I am not missing anything

parasol_cpu/src/test_utils.rs
```diff
@@ -166,4 +287,15 @@ mod tests {
             assert_eq!(val.get(&enc, &sk), i as u8);
         }
     }
+
+    #[test]
+    fn can_roundtrip_maybeint() {
+        let enc = get_encryption_80();
+        let sk = get_secret_keys_80();
+
+        for i in 118..138 {
```
cedricjiang, line 296: this crossed the positive/negative border for u8 so probably good?

rickwebiii (approved):

PR #46, opened 2025-05-05
Signed comparison in processor with testing

cedricjiang:


parasol_cpu/src/proc/assembly.rs
```diff
@@ -365,6 +365,18 @@ define_op! {
     // Compare less than or equal
     [0x43 CmpLe (dst dst, 0, Register) (src a, 0, Register) (src b, 0, Register)],
 
+    // Compare greater than, signed
+    [0x50 CmpGtS (dst dst, 0, Register) (src a, 0, Register) (src b, 0, Register)],
```
cedricjiang, line 369: Add S suffix in enum name, number is just 4->5
does not matter, we can always change if needed

cedricjiang: note there's no signed version for `CmpEq`, obviously

parasol_cpu/src/proc/ops/comparisons.rs
```diff
@@ -24,7 +33,7 @@ impl FheProcessor {
         b: RobEntryRef<Register>,
         instruction_id: usize,
         pc: u32,
-        operation: fn(u128, u128) -> bool,
+        operation: fn(u128, u32, u128, u32) -> bool,
```
cedricjiang, line 36: need width for signed

parasol_cpu/src/proc/tests/comparisons.rs
```diff
@@ -31,20 +31,39 @@ fn run_single_test(
...
-            encrypted_computation,
-        ))
-        .arg(MaybeEncryptedUInt::<32>::new(
-            val2 as u64,
-            enc,
-            &sk,
-            encrypted_computation,
-        ))
-        .return_value::<MaybeEncryptedUInt<32>>();
+    let args = if matches!(
```
cedricjiang, line 34: figure out signedness from op code

cedricjiang: I don't find a way to avoid repetition in the body of this if..else..

parasol_cpu/src/proc/tests/comparisons.rs
```diff
@@ -215,3 +234,107 @@ fn can_less_than_or_equal_ciphertext_inputs() {
         true,
     );
 }
+
+#[test]
+fn can_greater_than_signed_plaintext_inputs() {
+    run_comparison_test(
+        |val1, val2| val1 as i32 > val2 as i32,
```
cedricjiang, line 241: test values are actually 32bit integers, that is width == 32, so this "as" thing will just work

rickwebiii (approved):

PR #47, opened 2025-05-06
Rweber/aliases

rickwebiii:


parasol_cpu/src/proc/mod.rs
```diff
@@ -96,6 +99,112 @@ pub enum Register {
...
+        ($id:literal) => {
+            paste! {
+            #[doc = concat!("The `RegisterName` of ", stringify!([<X $id>]))]
+            pub const [<X $id>]: crate::tomasulo::registers::RegisterName<super::Register> =
+                    crate::tomasulo::registers::RegisterName::new($id);
+            }
+        };
+    }
+
+    def_name!(0);
```
cedricjiang, line 116: dunno if it's better to just do something like (if possible)

```rust
def_name!(0..64);
```

cedricjiang: hold on so it's 0 and 64 inclusive? so we have 65 registers?

parasol_cpu/src/proc/tests/add.rs
```diff
@@ -85,15 +79,9 @@ fn can_add_carry_inputs() {
...
-            IsaOp::AddC(
-                RegisterName::new(10),
-                RegisterName::new(1),
-                RegisterName::new(10),
-                RegisterName::new(11),
-                RegisterName::new(12),
-            ),
-            IsaOp::Zext(RegisterName::new(11), RegisterName::new(1), 32),
+            IsaOp::Trunc(A2, A2, 1),
+            IsaOp::AddC(A0, A1, A0, A1, A2),
```
cedricjiang, line 83: so both `RegisterName::new(1)` and `RegisterName::new(11)` map to `A1`? or you technically changed the original code (seems fine)

parasol_cpu/src/proc/tests/branch.rs
```diff
@@ -29,18 +29,10 @@ fn can_branch_zero() {
     // }
     // a
     let program = memory.allocate_program(&[
-        IsaOp::Add(
-            RegisterName::new(10),
-            RegisterName::new(10),
-            RegisterName::new(11),
-        ),
+        IsaOp::Add(A0, A0, A1),
         // Have we hit RegisterName::named(2)?
```
cedricjiang, line 33: what about this existing comment

parasol_cpu/src/proc/tests/add.rs
cedricjiang: looks like there are other places with this kind of change

rickwebiii: In some cases, I was arbitrarily using registers, so I can remap them to whatever I want.

cedricjiang (approved):

PR #51, opened 2025-05-07
Rweber/elf

rickwebiii:


parasol_cpu/src/proc/ops/casting.rs
```diff
@@ -137,14 +137,15 @@ impl FheProcessor {
...
         dst: RobEntryRef<Register>,
         src: RobEntryRef<Register>,
-        new_width: u32,
+        log_width: u32,
         instruction_id: usize,
         pc: u32,
     ) {
+        let new_width = 0x1u32 << log_width;
         let trunc_impl = || -> Result<()> {
             unwrap_registers!((mut dst) (src));
```
ryanorendorff, line 147: Does this change soon with the width being encoded as just bits?

ryanorendorff (approved):
Why did you check in add.o?

parasol_cpu/tests/test_data/add.o
cedricjiang, line 1: not sure if we want this file...my understanding is testing does not need this

parasol_cpu/src/proc/ops/casting.rs
cedricjiang: I think so, so merging https://github.com/Sunscreen-tech/hpu/pull/52 first helps

parasol_cpu/src/memory/mod.rs
```diff
@@ -218,10 +227,7 @@ impl Memory {
     /// `elf_data` should contain the contents of an executable ELF file generated by the Parasol
     /// linker.
     pub fn new_from_elf(elf_data: &[u8]) -> Result<Self> {
-        let memory = Self {
-            pages: Mutex::new(vec![None; TOTAL_PAGES as usize]),
-            stack_ptr: Mutex::new(Ptr32(0)),
-        };
+        let mut memory = Self::new(Ptr32(0xFFFF_FFFF - 16384), 16384);
```
cedricjiang, line 230: what's magical about 16384 (PS may write in hex?)

parasol_cpu/src/memory/mod.rs
```diff
@@ -275,6 +280,25 @@ impl Memory {
...
 
+        let (syms, sym_names) = elf.symbol_table()?.ok_or(Error::ElfNoSymbolTable)?;
+
+        // Create our symbol table.
+        for sym in syms {
+            if sym.st_symtype() != STT_FUNC {
+                continue;
+            }
+
+            let name = sym_names.get(sym.st_name as usize)?;
```
cedricjiang, line 291: pedantic but you only need name after you pass the check below

just saying

PR #52, opened 2025-05-07
Make width in instruction use number of bits to be consistent with compiler and add custom encoding as needed

cedricjiang:


parasol_cpu/src/proc/assembly.rs
```diff
@@ -13,6 +23,7 @@ macro_rules! define_op {
             $((dst $dst_name:ident, $dst_type_id:tt, $dst_type:ty))*
             $((src $src_name:ident, $src_type_id:tt, $src_type:ty))*
             $((meta $meta_name:ident, $meta_width:literal, $meta_type:ty))*
+            $((cmeta $cmeta_name:ident, $cmeta_width:literal, $cmeta_type:ty, $cmeta_dec:ident, $cmeta_enc:ident))*
```
cedricjiang, line 26: add this so we don't touch others, seems fine to me

I am not sure whether this means that if an instruction has both meta and cmeta then meta must come before cmeta, but not a restriction anyway

cedricjiang: c = complex

parasol_cpu/src/proc/assembly.rs
```diff
@@ -89,10 +101,17 @@ macro_rules! define_op {
                                         let value = value >> $meta_width;
                                     )*
 
+                                    $(
+                                        let mask = (0x1u64 << $cmeta_width) - 1;
+                                        let $cmeta_name = $cmeta_dec((value & mask) as $cmeta_type as u64) as $cmeta_type;
```
cedricjiang, line 106: decode here

parasol_cpu/src/proc/assembly.rs
```diff
@@ -129,6 +149,10 @@ macro_rules! define_op {
                                         let encoded = encoded | ($meta_name as u64) << shift;
                                         shift += $meta_width;
                                     )*
+                                    $(
+                                        let encoded = encoded | $cmeta_enc($cmeta_name as u64) << shift;
```
cedricjiang, line 153: encode here

parasol_cpu/src/proc/ops/mod.rs
```diff
@@ -119,8 +119,8 @@ pub fn make_parent_op(retirement_info: &RetirementInfo<DispatchIsaOp>) -> Arc<Co
     }))
 }
 
-pub(crate) fn is_invalid_load_store_alignment(base_addr: u32, width: u32) -> bool {
-    base_addr % width != 0 || width > 16 || !width.is_power_of_two() || width == 0
+pub(crate) fn is_invalid_load_store_alignment(base_addr: u32, num_bytes: u32) -> bool {
+    num_bytes == 0 || base_addr % num_bytes != 0 || num_bytes > 16 || !num_bytes.is_power_of_two()
```
cedricjiang, line 123: rename for clarify, also the zero check should come first, because `% 0` panics

ryanorendorff (approved):
I don't think the macro needs to take in the encoding and decoding functions; you could just specify specifically those in the macro body. 

The trunc/zext/sext also need to use this encoding I believe.

rickwebiii (approved):

cedricjiang:
Yes trunc and friends have them, as you can see in the code

```rust
    [0x05 Zext (dst dst, 0, Register) (src src, 0, Register) (cmeta width, 7, u32, width_dec, width_enc)],
    [0x06 Trunc (dst dst, 0, Register) (src src, 0, Register) (cmeta width, 7, u32, width_dec, width_enc)],
    [0x07 Sext (dst dst, 0, Register) (src src, 0, Register) (cmeta width, 7, u32, width_dec, width_enc)],
```

The reason for having external enc/dec functions is for later extensibility so if we want to encode / decode other information, but this may be a bit over-engineering

PR #56, opened 2025-05-08
Rweber/cardio

rickwebiii:


parasol_cpu/src/proc/assembly.rs
```diff
@@ -346,7 +346,7 @@ define_op! {
     [0x02 Load (dst dst, 0, Register) (src src, 0, Register) (cmeta width, 7, u32, width_dec, width_enc)],
 
     // Load immediate
-    [0x03 LoadI (dst dst, 0, Register) (meta imm, 32, u128) (cmeta width, 7, u32, width_dec, width_enc)],
+    [0x03 LoadI (dst dst, 0, Register) (meta imm, 32, i32) (cmeta width, 7, u32, width_dec, width_enc)],
```
cedricjiang, line 349: why i32?

cedricjiang: ok so immediate numbers must be signed 32 bit number?

rickwebiii: 1. The immediate is 32-bits
2. Our compiler emits `loadi -90, 8`.
3. Loading negative values is a sensible thing to do.

cedricjiang: the compiler technically does not emit `loadi -90, 8`, it emits bits that the disassembler somehow decides to show `-90` for the bits that correspond to `imm`

parasol_cpu/src/proc/tests/call_abi.rs
```diff
@@ -127,8 +127,8 @@ fn large_return_value() {
     let memory = Arc::new(Memory::new_default_stack());
 
     let program = memory.allocate_program(&[
-        IsaOp::LoadI(T0, 0xDEADBEEF, 32),
-        IsaOp::LoadI(T1, 0xFEEDF00D, 32),
+        IsaOp::LoadI(T0, 0xDEADBEEFu32 as i32, 32),
```
cedricjiang, line 130: 0xDEADBEEFi32 should just work...

actually because `LoadI` enum's second data field if `i32`, I _think_ you can probably just keep it as is

rickwebiii: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&code=fn+main%28%29+%7B%0A++++let+x+%3D+0xDEADBEEFi32%3B%0A%7D

```
error: literal out of range for `i32`
 --> src/main.rs:2:13
  |
2 |     let x = 0xDEADBEEFi32;
  |             ^^^^^^^^^^^^^
  |
  = note: the literal `0xDEADBEEFi32` (decimal `3735928559`) does not fit into the type `i32` and will become `-559038737i32`
  = note: `#[deny(overflowing_literals)]` on by default
help: consider using the type `u32` instead
  |
2 -     let x = 0xDEADBEEFi32;
2 +     let x = 0xDEADBEEFu32;
  |
help: to use as a negative number (decimal `-559038737`), consider using the type `u32` for the literal and cast it to `i32`
  |
2 -     let x = 0xDEADBEEFi32;
2 +     let x = 0xDEADBEEFu32 as i32;
  |             ^^^^^^^^^^^^^
  ```

You need the explicit cast and this is

PR #59, opened 2025-05-08
Rweber/cleanup

rickwebiii:


parasol_cpu/tests/test_src/cmux/cmux.c
```diff
@@ -8,4 +8,3 @@
 ) {
     return bound > 10 ? a : b;
 }
```
rickwebiii, line 10: Revert. C99 requires a newline at EOF

rickwebiii: Nevermind, the diff is saying I had an extra \n

PR #60, opened 2025-05-08
run_program no longer meters, add run_program_metered

ryanorendorff:
Metering is used more for services that deploy the HPU, versus just running the programs locally. We use `run_program` without any meaningful gas limit far more than we run it with one. Hence this PR makes the default to have no gas limit at all and adds a `run_program_metered` that does meter the processor. 

Note that one can run `run_program_metered` with a `None` limit to get out the amount of gas used by a computation without specifying a gas limit.

### DIFF
--- a/parasol_cpu/src/proc/fhe_processor.rs
+++ b/parasol_cpu/src/proc/fhe_processor.rs
@@ -222,7 +222,7 @@ impl FheProcessor {
         &mut self,
         inst: IsaOp,
         pc: u32,
-        gas_limit: u32,
+        gas_limit: Option<u32>,
     ) -> Result<(u32, u32)> {
         use crate::tomasulo::{GetDeps, ToDispatchedOp};
 
@@ -249,8 +249,10 @@ impl FheProcessor {
 
         let gas = self.compute_gas(&disp_inst);
 
-        if gas > gas_limit {
-            return Err(Error::OutOfGas(gas, gas_limit));
+        if let Some(gas_limit) = gas_limit {
+            if gas > gas_limit {
+                return Err(Error::OutOfGas(gas, gas_limit));
+            }
         }
 
         scoreboard_entry.set_instruction(&disp_inst);
@@ -666,13 +668,14 @@ impl FheProcessor {
     }
 
     /// Runs the given program using the passed user `data` as arguments with a gas limit
-    /// Returns used gas a program return value
-    pub fn run_program<T: ToArg>(
+    /// Returns the amount of gas used to run the program and the program return
+    /// value
+    pub fn run_program_metered<T: ToArg>(
         &mut self,
         memory: &Arc<Memory>,
         initial_pc: Ptr32,
         args: &Args<T>,
-        gas_limit: u32,
+        gas_limit: Option<u32>,
     ) -> Result<(u32, T)> {
         self.reset()?;
         let return_data = self.set_up_function_call(memory, args)?;
@@ -697,7 +700,14 @@ impl FheProcessor {
                         Error::Halt => break,
                         Error::OutOfGas(used_gas, _) => {
                             self.wait()?;
-                            return Err(Error::OutOfGas(gas + used_gas, gas_limit));
+                            if let Some(gas_limit) = gas_limit {
+                                return Err(Error::OutOfGas(gas + used_gas, gas_limit));
+                            } else {
+                                // This case should never happen since gas
+                                // tracking should throw an out of gas answer
+                                // only when the gas_limit is a Some variant.
+                                unreachable!()
+                            }
                         }
                         _ => return Err(e),
                     },
@@ -717,6 +727,18 @@ impl FheProcessor {
 
         self.try_capture_return_value(memory, args, return_data)
             .map(|ret_val| (gas, ret_val))
+    }
+
+    /// Runs the given program using the passed user `data` as arguments.
+    /// Returns the result of the program.
+    pub fn run_program<T: ToArg>(
+        &mut self,
+        memory: &Arc<Memory>,
+        initial_pc: Ptr32,
+        args: &Args<T>,
+    ) -> Result<T> {
+        self.run_program_metered(memory, initial_pc, args, None)
+            .map(|x| x.1)
     }
 }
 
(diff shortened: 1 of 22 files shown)


rickwebiii (approved):

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -706,7 +709,14 @@ impl FheProcessor {
                         Error::Halt => break,
                         Error::OutOfGas(used_gas, _) => {
                             self.wait()?;
-                            return Err(Error::OutOfGas(gas + used_gas, gas_limit));
+                            if let Some(gas_limit) = gas_limit {
```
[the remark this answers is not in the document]
rickwebiii: There's a clear comment right above it stating why this can never be hit. Your suggestion is bad and you should feel bad.

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -727,6 +737,18 @@ impl FheProcessor {
...
+    /// Runs the given program using the passed user `data` as arguments.
+    /// Returns the result of the program.
+    pub fn run_program<T: ToArg>(
+        &mut self,
+        memory: &Arc<Memory>,
+        initial_pc: Ptr32,
+        args: &Args<T>,
+    ) -> Result<T> {
+        self.run_program_metered(memory, initial_pc, args, None)
+            .map(|x| x.1)
```
cedricjiang, line 750: I know this is probably not idiomatic rust, but if you put in u32::MAX you can avoid all these option boilerplates

cedricjiang: just saying

parasol_cpu/src/proc/tests/add.rs
```diff
@@ -26,9 +26,7 @@ fn can_add_inputs() {
             .arg(MaybeEncryptedUInt::<32>::new(val2 as u64, &enc, &sk, enc2))
             .return_value::<MaybeEncryptedUInt<32>>();
 
-        let (_, result) = proc
-            .run_program(program, &Arc::new(memory), args, 200_000)
-            .unwrap();
+        let result = proc.run_program(program, &Arc::new(memory), args).unwrap();
```
cedricjiang, line 29: I actually intentionally put the gas limit there to get a sense of their consumption, but ok...

parasol_cpu/src/proc/fhe_processor.rs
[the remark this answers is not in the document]
cedricjiang: why do you fight with AI...

parasol_cpu/src/proc/fhe_processor.rs
rickwebiii: At 100k gas per encrypted operation, assuming 1s per operation (wildly conservative), 4e9 gas gets used up in 11 hours. If someone wants to disable gas and run a computation for a day, we should let them.

rickwebiii: Nothwithstanding my disdain for sentinel values

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -258,8 +258,10 @@ impl FheProcessor {
 
         let gas = self.compute_gas(&disp_inst);
 
-        if gas > gas_limit {
-            return Err(Error::OutOfGas(gas, gas_limit));
+        if let Some(gas_limit) = gas_limit {
```
cedricjiang, line 261: I think there's something like
 ```rust
if gas_limit.map_or(false, |v| gas > v) {
   return Err(..);
}
```

File: parasol_cpu/src/proc/fhe_processor.rs (hunk 2 above)
Comment (cedricjiang) [reply parasol_cpu/src/proc/fhe_processor.rs:750]: time to use `u64` gas :)

Comment (ryanorendorff) [reply parasol_cpu/src/proc/fhe_processor.rs:750]: We should let people run indefinitely. Also sentinel values make Ryan sad.

File: parasol_cpu/src/proc/tests/add.rs (hunk 3 above)
Comment (ryanorendorff) [reply parasol_cpu/src/proc/tests/add.rs:29]: We should add in a benchmark so that we get a consistent amount of gas per time on a given machine, instead of pulling the value out of a test.

File: parasol_cpu/src/proc/fhe_processor.rs (hunk 4 above)
Comment (ryanorendorff) [reply parasol_cpu/src/proc/fhe_processor.rs:261]: I believe the more idiomatic thing here is to use a if let.

Comment (cedricjiang) [reply parasol_cpu/src/proc/fhe_processor.rs:261]: dunno, I have a lot of ways to write this in my brain but I don't like any of them, including what I said above...

File: parasol_cpu/src/proc/fhe_processor.rs (hunk 2 above)
Comment (cedricjiang) [reply parasol_cpu/src/proc/fhe_processor.rs:750]: thats's fine...but in reality I don't think we will actually let people run indefinitely :)

File: parasol_cpu/src/proc/tests/add.rs (hunk 3 above)
Comment (cedricjiang) [reply parasol_cpu/src/proc/tests/add.rs:29]: that's why we have https://github.com/Sunscreen-tech/hpu/issues/36

File: parasol_cpu/src/proc/fhe_processor.rs [hunk 5]
```diff
@@ -19,6 +19,50 @@ use std::sync::{
...
+}
+
+impl RunProgramOptionsBuilder {
+    /// Creates a new [`RunProgramOptionsBuilder`]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Set the gas limit.
+    pub fn gas_limit(mut self, gas_limit: Option<u32>) -> Self {
```
cedricjiang, line 53: do we expect a case where someone calls this function but passes in a `None`?

ryanorendorff: If they specifically want to run without a gas limit for sure.

cedricjiang: well that's the default...anyway

ryanorendorff: You could be picking up the value from someone else (like a config file) where someone has explicitly said none. Or we could change the default. We should allow people to be explicit with their choice if they want.

cedricjiang: yeah that's fine

PR #72, opened 2025-05-17
Make opcodes consistent with compiler as of now

cedricjiang:


cedricjiang:
Also moved the order so they are sorted for now

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -882,6 +883,9 @@ impl Tomasulo for FheProcessor {
                     pc,
                 );
             }
+            Move(..) => {
+                unimplemented!()
```
cedricjiang, line 887: @rickwebiii for you :)

PR #73, opened 2025-05-19
Make word size a constant

ryanorendorff:
The point of this PR is to change the magic values of `4` and `8` into meanings based on the word size of the processor. I may have missed parts, but at least this is a start.

### DIFF
--- a/parasol_cpu/src/memory/mod.rs
+++ b/parasol_cpu/src/memory/mod.rs
@@ -31,6 +31,12 @@ const PAGE_ID_MASK: u32 = !PAGE_OFFSET_MASK;
 // The total number of pages that fit in the 32-bit address space.
 const TOTAL_PAGES: u32 = 0x1 << (32 - LOG2_PAGE_SIZE);
 
+// Bytes in a word
+pub(crate) const WORD_SIZE: u32 = 4;
+
+// Number of bytes in an instruction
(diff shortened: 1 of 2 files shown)


parasol_cpu/src/memory/mod.rs
```diff
@@ -440,36 +446,39 @@ impl Memory {
...
-        Ok(u32::from_le_bytes([b0, b1, b2, b3]))
+        let bytes = self.try_load_n_bytes::<{ WORD_SIZE as usize }>(virtual_address)?;
+        Ok(u32::from_le_bytes(bytes))
     }
 
-    /// Attempt to load a 4-byte plaintext word from the given address.
+    /// Attempt to load a plaintext double word from the given address.
     pub(crate) fn try_load_plaintext_dword(&self, virtual_address: Ptr32) -> Result<u64> {
-        if virtual_address.0 % 8 != 0 {
+        if virtual_address.0 % (2 * WORD_SIZE) != 0 {
```
cedricjiang, line 476: you can have a `DWORD_SIZE` constant maybe

cedricjiang: looks like there could be quite a bikeshed whether these constants should be u32 or usize :)

ryanorendorff: Sure

PR #74, opened 2025-05-19
Add move implementation with testing

cedricjiang:


parasol_cpu/src/proc/ops/mov.rs
```diff
@@ -0,0 +1,45 @@
...
+                        val: *val,
+                        width: *width,
+                    };
+                }
+                Register::Ciphertext(ciphertext) => {
+                    *dst = Register::Ciphertext(match ciphertext {
+                        Ciphertext::L0Lwe { data } => Ciphertext::L0Lwe { data: data.clone() },
+                        Ciphertext::L1Lwe { data } => Ciphertext::L1Lwe { data: data.clone() },
+                        Ciphertext::L1Glwe { data } => Ciphertext::L1Glwe { data: data.clone() },
+                        Ciphertext::L1Ggsw { data } => Ciphertext::L1Ggsw { data: data.clone() },
```
cedricjiang, line 31: bummer but...well

parasol_cpu/src/proc/tests/mov.rs
```diff
@@ -0,0 +1,53 @@
...
+        assert_eq!(
+            src_val, ret_val,
+            "src_val: {:#02x}, ans_sum: {:#02x}, [unimportant from here] src_enc: {}, dst_val: {:#02x}, dst_enc; {}",
+            src_val, ret_val, src_enc, dst_val, dst_enc
+        );
+    };
+
+    for src_enc in [true, false] {
+        for dst_enc in [true, false] {
+            test(0xF00Du32, src_enc, 0xFEEDu32, dst_enc);
```
cedricjiang, line 50: src value is the only thing that matters but would like to test all 4 combinations anyway

parasol_cpu/src/proc/ops/mov.rs
ryanorendorff: It is what it is

ryanorendorff (approved):

parasol_cpu/src/proc/ops/mov.rs
cedricjiang: I wonder if we should derive clone for Ciphertext but maybe not yet

PR #77, opened 2025-05-19
Update test program source and binary with latest compiler

cedricjiang:


ryanorendorff (approved):
Looks good. If you want, add a Make script that compiles everything so this is automated next time.

PR #90, opened 2025-05-26
Create a "dynamic" version of fluent types and argument conversion that do not require a constant size for flexibility

cedricjiang:


parasol_cpu/src/proc/args.rs
```diff
@@ -75,6 +75,71 @@ where
...
+        self.bytes_to_arg(bytes, Self::is_signed())
+    }
+
+    /// Describe this type when it appears in the return value of a function call.
+    /// This allows the Parasol processor to understand how to capture the return value from a
+    /// program.
+    ///
+    /// See [here](https://drive.google.com/file/d/1Ja_Tpp_5Me583CGVD-BIZMlgGBnlKU4R/view?pli=1) for
+    /// details.
+    fn to_return_value(&self) -> ReturnValue<Self> {
```
cedricjiang, line 131: I probably should delete this, since you cannot really use a `DynamicToArg` for return value because itself does not include any size information

parasol_runtime/src/fluent/int.rs
```diff
@@ -139,6 +208,24 @@ mod tests {
...
+    #[test]
+    fn can_unsafe_deserialize_dyn_int() {
+        fn case<T: CiphertextOps + for<'a> Deserialize<'a> + Serialize>() {
+            let enc = get_encryption_128();
+            let sk = get_secret_keys_128();
+
+            let val = DynamicInt::<T>::encrypt_secret(2u64.pow(15) - 42, &enc, &sk, 15);
+
+            let ser = bincode::serialize(&val).unwrap();
+            bincode::deserialize::<DynamicInt<T>>(&ser).unwrap();
```
cedricjiang, line 220: note because there's really no way to let `DynamicInt` implement `GetSize`, we have to do "unsafe" bincode deserialization here (same for uint of course)

parasol_cpu/src/proc/args.rs
```diff
@@ -75,6 +75,71 @@ where
...
+/// Similar to [`ToArg`] but the alignment and size are figured out from the instance not just the type
+pub trait DynamicToArg
+where
+    Self: Sized,
+{
+    /// The required alignment for this type. Must be power of 2 and should be no larger than 4096.
+    fn alignment(&self) -> usize;
+
+    /// The number of bytes this type takes. Should be no greater than `u32::MAX`.
+    fn size(&self) -> usize;
```
rickwebiii, line 87: Just update the existing trait to be this.

parasol_cpu/src/proc/args.rs
```diff
@@ -324,6 +389,82 @@ impl ToArg for () {
     }
 }
 
+impl DynamicToArg for DynamicUInt<L1GlweCiphertext> {
```
rickwebiii, line 392: This thing isn't necessarily an integer as users could put a struct, array, or whatever in here.

parasol_cpu/src/proc/args.rs
```diff
@@ -324,6 +389,82 @@ impl ToArg for () {
...
+            .collect::<Result<Vec<_>>>()?
+            .into_iter()
+            .flatten()
+            .collect::<Vec<_>>();
+
+        Ok(DynamicUInt::from_bits_shallow(data))
+    }
+}
+
+impl DynamicToArg for DynamicInt<L1GlweCiphertext> {
```
rickwebiii, line 430: I think we can drop this type.

parasol_cpu/src/proc/args.rs
rickwebiii: Do we have a way for users to indicate pass by reference vs value?

parasol_cpu/src/proc/mod.rs
```diff
@@ -312,6 +315,23 @@ impl FheComputer {
         packed_ct
     }
 
+    /// Similar to [`FheComputer::pack_int`] but works on [`DynamicGenericInt`]
+    pub fn pack_int_dyn<U: Sign>(
```
rickwebiii, line 319: Non-sign extended version?

parasol_cpu/src/proc/args.rs
cedricjiang: I tried that...it fails to work on some existing stuff which requires the alignment/size to be known just because of the type, not dependent on the instance.

parasol_cpu/src/proc/args.rs
```diff
@@ -324,6 +389,82 @@ impl ToArg for () {
...
+    fn size(&self) -> usize {
+        self.bits.len() / 8
+    }
+
+    fn is_signed() -> bool {
+        false
+    }
+
+    fn to_bytes(&self) -> Vec<Byte> {
+        assert!(self.bits.len().is_power_of_two() && self.bits.len() % 8 == 0);
```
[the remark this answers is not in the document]
cedricjiang: Let me think about this, I think the caller into this function should have run the check and returned to the user if this assertion does not hold

parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -515,6 +515,148 @@ impl<'a, const N: usize, V: Sign> GenericIntGraphNodes<'a, N, L1GgswCiphertext,
...
+    /// The generic integer's [`BitNode`]s from least to most significant.
+    pub bits: &'a [BitNode<T>],
+
+    _phantom: PhantomData<U>,
+}
+
+impl<'a, T: CiphertextOps, U: Sign> DynamicGenericIntGraphNodes<'a, T, U> {
+    pub(crate) fn from_nodes<I: Iterator<Item = NodeIndex>>(
+        iter: I,
+        n: usize,
```
rickwebiii, line 529: See comment below about `ExactSizeIterator`

parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -515,6 +515,148 @@ impl<'a, const N: usize, V: Sign> GenericIntGraphNodes<'a, N, L1GgswCiphertext,
...
+    }
+}
+
+impl<U: Sign> DynamicGenericIntGraphNodes<'_, L1GlweCiphertext, U> {
+    /// Similar to [`GenericIntGraphNodes::pack`] but works on [`DynamicGenericIntGraphNodes`]
+    pub fn pack(
+        &self,
+        ctx: &FheCircuitCtx,
+        enc: &Encryption,
+        n: usize,
```
rickwebiii, line 597: Don't need n. See above.

parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -577,6 +719,55 @@ impl<const N: usize, T: CiphertextOps + PolynomialCiphertextOps, U: Sign>
...
+    id: NodeIndex,
+    _phantom: PhantomData<(T, U)>,
+}
+
+impl<V: Sign> PackedDynamicGenericIntGraphNode<L1GlweCiphertext, V> {
+    /// Similar to [`PackedGenericIntGraphNode::unpack`] but works on [`PackedDynamicGenericIntGraphNode`]
+    pub fn unpack<'a>(
+        &self,
+        ctx: &'a FheCircuitCtx,
+        n: usize,
```
rickwebiii, line 733: Ideally, this isn't exposed publicly. Put `n` in the type and let `DynamicGenericInt` populate it when you call `graph_inputs()`

parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -577,6 +719,55 @@ impl<const N: usize, T: CiphertextOps + PolynomialCiphertextOps, U: Sign>
...
+    }
+}
+
+impl<T: CiphertextOps + PolynomialCiphertextOps, U: Sign> PackedDynamicGenericIntGraphNode<T, U> {
+    /// Similar to [`PackedGenericIntGraphNode::collect_output`] but works on [`PackedDynamicGenericIntGraphNode`]
+    pub fn collect_output(
+        &self,
+        ctx: &FheCircuitCtx,
+        enc: &Encryption,
+        n: usize,
```
rickwebiii, line 754: After moving `n` into the type, you can remove this.

parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -714,6 +905,87 @@ where
     }
 }
 
+#[derive(Clone, Serialize, Deserialize)]
+/// Similar to [`GenericInt`] but without the size N generic parameter
+pub struct DynamicGenericInt<T: CiphertextOps, U: Sign> {
```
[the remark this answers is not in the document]
rickwebiii: Our calling convention just ignores zero-sized types in call signatures, so maybe you can allow those?

parasol_runtime/src/fluent/generic_int.rs
cedricjiang: good to know...one reason I pass `n` is because of you don't get size from iter

parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -714,6 +905,87 @@ where
...
+    _phantom: PhantomData<U>,
+}
+
+impl<T, U> DynamicGenericInt<T, U>
+where
+    T: CiphertextOps,
+    U: Sign,
+{
+    /// Similar to [`GenericInt::new`] but generates [`DynamicGenericInt`]
+    pub fn new(enc: &Encryption, n: usize) -> Self {
```
rickwebiii, line 922: I think this `n` is okay.

parasol_runtime/src/fluent/generic_int.rs
cedricjiang: what does "Put `n` in the type" mean? if you mean generic param then it goes back to `GenericInt` which stops us from having flexible sizes.

parasol_runtime/src/fluent/generic_int.rs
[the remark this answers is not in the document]
cedricjiang: that's likely in another repo, but yes, technically the parasol runtime can accept arbitrary bit count / size here (just like the version with `const N: usize`)

parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -714,6 +905,87 @@ where
...
 
+#[derive(Clone, Serialize, Deserialize)]
+/// Similar to [`GenericInt`] but without the size N generic parameter
+pub struct DynamicGenericInt<T: CiphertextOps, U: Sign> {
+    /// The ciphertexts encrypting this generic integer's bits in least-to-most significant order.
+    pub bits: Vec<Arc<AtomicRefCell<T>>>,
+    _phantom: PhantomData<U>,
+}
+
+impl<T, U> DynamicGenericInt<T, U>
```
rickwebiii, line 916: Idea for another CR, migrate the existing `GenericInt<N>` types be a simple wrapper around this type that passes `N` for `n`. This removes code duplication.

parasol_cpu/src/proc/mod.rs
rickwebiii: Yeah, there's no reason for packing to worry about sign extension.

parasol_cpu/src/proc/args.rs
rickwebiii: Sorry, just thinking aloud for the HPU service, which is a different repo.

rickwebiii: The HPU supports either.

parasol_cpu/src/proc/args.rs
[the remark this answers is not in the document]
rickwebiii: The HPU service can gatekeep anything it deserializes and abort.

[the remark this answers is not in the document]
cedricjiang: yeah so we can keep the assertion here and let the "front end" deal with wrong user inputs

parasol_cpu/src/proc/args.rs
cedricjiang: there is support for dynamic return values, but it's not using this specific function

parasol_cpu/src/proc/args.rs
cedricjiang: I will figure out why `ToArg` fails in one of my attempts

parasol_cpu/src/proc/args.rs
cedricjiang: You mea there is pass by reference? I wonder how that's done.

parasol_cpu/src/proc/args.rs
cedricjiang: sorry I think I misunderstood. in `ToArg` they are already functions instead of constants (done in https://github.com/Sunscreen-tech/hpu/pull/87), but they do not take `&self` which is needed here. When I say "I tried that", I mean I tried to make `&self` arguments for these functions in `ToArg` but what fails is then you cannot really write `try_from_bytes` for `impl<const N: usize, T: ToArg> ToArg for [T; N]`

parasol_cpu/src/proc/args.rs
```diff
@@ -75,6 +75,71 @@ where
...
+
+    /// The number of bytes this type takes. Should be no greater than `u32::MAX`.
+    fn size(&self) -> usize;
+
+    /// Whether this type needs to be sign extended or not.
+    fn is_signed() -> bool;
+
+    /// Convert a byte array and metadata into an [`Arg`] indicating the alignment,
+    /// size, and how to extend the given value.
+    fn bytes_to_arg(&self, bytes: Vec<Byte>, is_signed: bool) -> Arg {
```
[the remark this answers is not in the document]
cedricjiang: deleted for now

parasol_cpu/src/proc/args.rs
cedricjiang: deleted for now

parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -515,6 +515,148 @@ impl<'a, const N: usize, V: Sign> GenericIntGraphNodes<'a, N, L1GgswCiphertext,
...
+
+        Self {
+            bits: nodes,
+            _phantom: PhantomData,
+        }
+    }
+
+    pub(crate) fn from_bit_nodes<I: Iterator<Item = BitNode<T>>>(
+        iter: I,
+        n: usize,
```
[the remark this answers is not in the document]
cedricjiang: yeah this actually help getting rid of all `n` params for the pack/unpack function

parasol_cpu/src/proc/args.rs
cedricjiang: need to think a bit more, but for parity with `ToArg` I think I will want to keep this for now

parasol_runtime/src/fluent/generic_int.rs
cedricjiang: yes

parasol_runtime/src/fluent/generic_int.rs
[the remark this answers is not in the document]
cedricjiang: what's the comment about zero sized types about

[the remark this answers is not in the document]
cedricjiang: I think when you mention calling convention you mean for parasol program which is C for now?

PR #93, opened 2025-05-29
Wrap fluent integer types with constant size generic over the dynamic ones

cedricjiang:


parasol_runtime/src/fluent/int.rs
```diff
@@ -107,6 +108,7 @@ mod tests {
         let ctx = FheCircuitCtx::new();
 
         let as_unpacked = val
+            .inner_ref()
```
rickwebiii, line 111: What's this inner_ref? I think if you impl Deref you can kill this thing.

cedricjiang: done...I wonder if there are ways to actually get rid of `inner` as well, maybe `Into`?

cedricjiang: ok solved

PR #107, opened 2025-06-12
Rweber/harden

rickwebiii:


parasol_cpu/src/error.rs
```diff
@@ -6,6 +6,11 @@ use thiserror::Error;
 #[derive(Debug, Clone, Error)]
 /// Errors that can occur in this crate.
 pub enum Error {
+    /// An error occurred when processing a running a circuit on a
```
cedricjiang, line 9: `a running circuit`

parasol_runtime/src/circuit_processor/task.rs
```diff
@@ -0,0 +1,213 @@
...
+                | (FheOp::MultiplyGgswGlwe, FheEdge::Ggsw) => ct.is_ggsw1(),
+                _ => unreachable!(),
+            };
+
+        if !is_valid {
+            return Err(RuntimeError::invalid_ciphertext_kind(&self, ct, edge));
+        }
+
+        Ok(())
+    }
```
cedricjiang, line 80: maybe just
```
if is_valid {
    Ok(())
} else {
    Err(...)
}

File: parasol_runtime/src/circuit_processor/task.rs [hunk 3]
```diff
@@ -0,0 +1,213 @@
...
+    pub fn validate(&self, params: &Params) -> Result<(), RuntimeError> {
+        self.validate_inputs()?;
+        self.validate_op(params)?;
+
+        Ok(())
+    }
+
+    #[inline]
+    fn validate_op(&self, params: &Params) -> Result<(), RuntimeError> {
+        match &self.op {
```
cedricjiang, line 35: can one arm match be just if let?

rickwebiii: I like match's more systematic vibes, as it's just by chance that this is the only op that requires validation. Future ones can just add a match variant.

parasol_runtime/src/circuit_processor/task.rs
cedricjiang: still inverted boolean flag though :)

cedricjiang: I am a bit surprised clippy is ok with this

parasol_runtime/src/circuit_processor/task.rs
cedricjiang: lol clippy asks for it...

PR #108, opened 2025-06-12
round during mod switch

rickwebiii:


ryanorendorff (approved):
Looks good. Would be good to have basic tests to make sure the round is rounding but otherwise it's good.

PR #109, opened 2025-06-13
Rename HPU->SPF

ryanorendorff:


mux_circuits/Cargo.toml
```diff
@@ -7,7 +7,7 @@ rust-version = "1.85.0"
 license = "AGPL-3.0-only"
 description = "This crate contains the Sunscreen Torus FHE (TFHE) implementation"
 homepage = "https://sunscreen.tech"
-repository = "https://github.com/Sunscreen-tech/Sunscreen"
+repository = "https://github.com/Sunscreen-tech/spf"
```
cedricjiang, line 10: this is the old BFV stuff right...didn't realize we still have reference to them in this repo...

ryanorendorff: Apparently so.

PR #114, opened 2025-06-23
Rweber/cc changes

rickwebiii:
* Change our calling convention to the following:
  * Allocate space of the stack for a function's arguments (in reverse order) and return value. This includes padding inserted as needed to accommodate each argument and the return value's alignment requirements as well as ensuring the resultant stack pointer is 16-byte aligned. Hence, SP+0 will be the first argument SP+arg_1_size will be the second and so on.
  * Subtract the current function's stack pointer (or, just the stack pointer if it's the entry point) by the allocation amount. Parasol stacks grow down.
  * Update the SP register to point to the new top of the stack.
  * Update RP to the return value's allocated space on the stack. RP always contains a pointer regardless of the underlying data's size.
* Add a debug instruction that, when debug handlers are installed, allow users to inspect register contents.
* Add an offset field to load/store (using LoadI all the time was too annoying). These changes are ISA compatible with our compiler, which will always emit 0 offsets.


parasol_cpu/src/proc/args.rs
```diff
@@ -450,29 +425,7 @@ impl DynamicToArg for DynamicInt<L1GlweCiphertext> {
...
-///
-/// 5-8 byte aggregate values get packed into 2 registers (if available). If
-/// only one register is available, the lo 32-bits go in the register while the high
-/// bits go on the stack. If no registers are available, both words get pushed on
-/// the stack.
-///
-/// Scalars or aggregates larger than 64-bits will be transparently allocated on the
-/// Parasol heap and passed by reference (i.e. a 32-bit pointer will be passed for the corresponding
-/// register/stack argument).
+/// Parasol's calling convention is similar to x86's cdecl
```
rickwebiii, line 428: Elaborate.

parasol_cpu/benches/cardio.rs
```diff
@@ -118,6 +118,13 @@ pub fn cardio_test_program() -> Vec<IsaOp> {
     let shift_amount = X32;
 
     vec![
+        IsaOp::Load(flags, SP, 8, 0),
+        IsaOp::Load(age, SP, 8, 1),
+        IsaOp::Load(hdl, SP, 8, 2),
+        IsaOp::Load(weight, SP, 8, 3),
+        IsaOp::Load(height, SP, 8, 4),
+        IsaOp::Load(physical_activity, SP, 8, 5),
+        IsaOp::Load(glasses_alcohol, SP, 8, 6),
```
cedricjiang, line 127: I wonder if we need `push` instruction (and thus `pop`)...probably will complicate stuff?

parasol_cpu/src/proc/args.rs
```diff
@@ -450,29 +425,14 @@ impl DynamicToArg for DynamicInt<L1GlweCiphertext> {
...
+/// storing all arguments and the return value as well as padding to maintain alignment
+/// requirements for each argument and the return value. Furthermore, the stack pointer must be
+/// 16-byte aligned after pushing all the call data. Arguments are stored in reverse order
+/// (i.e. the first argument appears at SP+0 while the second at SP+sizeof(arg1), and so-on),
+/// followed by the return value.
 ///
-/// Scalars or aggregates larger than 64-bits will be transparently allocated on the
-/// Parasol heap and passed by reference (i.e. a 32-bit pointer will be passed for the corresponding
-/// register/stack argument).
+/// Parasol stacks grow downwards, while arguments and values within a frame grow upwards.
```
cedricjiang, line 435: what does "arguments and values within a frame grow upwards" exactly mean?

parasol_cpu/src/proc/args.rs
```diff
@@ -558,69 +515,32 @@ pub struct ReturnValue<T> {
 }
 
 /// Arguments passed to an FHE program when calling [`crate::FheComputer::run_program`].
-pub struct Args<T> {
+pub struct CallData<T> {
```
cedricjiang, line 518: I wonder why the rename, also if you do so, do you need to change the doc above?

parasol_cpu/src/proc/args.rs
```diff
@@ -558,69 +515,32 @@ pub struct ReturnValue<T> {
...
+        }
+
+        if self.return_value.size > 0 {
+            ptr += (self.return_value.alignment - ptr % self.return_value.alignment)
+                % self.return_value.alignment;
+            ptr += self.return_value.size;
         }
 
-        (stack_size.next_multiple_of(16) - stack_size) as u32
+        ptr += (16 - ptr % 16) % 16;
```
cedricjiang, line 541: I wonder if this is just `next_multiple_of`

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -156,12 +170,12 @@ impl FheProcessor {
 
                     let num_bytes = width / 8;
 
+                    let base_addr = Ptr32::from(base_addr).try_signed_offset(offset)?;
```
cedricjiang, line 173: could inline above `base_addr`...just saying

parasol_cpu/src/proc/ops/dbg.rs
```diff
@@ -0,0 +1,27 @@
...
+        &self,
+        retirement_info: &RetirementInfo<DispatchIsaOp>,
+        src: RobEntryRef<Register>,
+        handler_id: u32,
+        instruction_id: usize,
+        pc: u32,
+    ) {
+        let handler_id = handler_id as usize;
+
+        if self.debug_handlers.len() > handler_id {
```
cedricjiang, line 19: I wonder if using `get` method to obtain an option (then `if let`) is more idiomatic, just saying...

PR #116, opened 2025-06-26
Fix bound checking in generic int

cedricjiang:


parasol_runtime/src/fluent/generic_int.rs
```diff
@@ -736,9 +736,9 @@ where
     /// Encrypts the given integer.
     ///
     /// # Panics
-    /// If `val > 2^N`
+    /// If `val >= 2^n` (only when `n` is 63 or smaller)
     pub fn encrypt_secret(val: u64, enc: &Encryption, sk: &SecretKey, n: usize) -> Self {
-        if n < 64 && val > 0x1 << n {
+        if n < 64 && val >= 0x1 << n {
```
cedricjiang, line 741: I think we need `>=` here, same below

PR #118, opened 2025-06-30
Make use of one step compilation and linking

cedricjiang:


parasol_cpu/tests/test_data/.gitignore
```diff
@@ -1,2 +0,0 @@
-# Allow for C object files since we use them for testing here
-*.o
```
cedricjiang, line 2: we no longer generate `.o` files, that said, this is already covered by the root `.gitignore` anyway...

PR #119, opened 2025-06-30
Fixes for 128-bit ints

rickwebiii:


ryanorendorff (approved):

parasol_cpu/benches/auction.rs
```diff
@@ -42,7 +43,7 @@ fn generate_args<const N: usize>(
     enc: &Encryption,
     sk: &SecretKey,
 ) -> CallData<[UInt<16, L1GlweCiphertext>; 2]> {
```
cedricjiang, line 45: you did not replace this one and the other to the alias

cedricjiang: similar in other files

PR #126, opened 2025-07-10
Rweber/faults

rickwebiii:


parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -468,14 +466,9 @@ impl FheProcessor {
...
-            // Waiting thread may have dropped.
-            let _ = retirement_info
-                .ready_instructions
-                .send(InstructionOperation::Retire(Err(e)));
-            return;
-        }
-
+        // We always notify our dependencies regardless of whether we errored or not.
+        // After doing that, we retire this instruction with success/failure as
+        // appropriate.
```
rickwebiii, line 471: We should set fault here rather than above so our dependencies will move to no-op mode rather than racing to go to no-op mode.

PR #130, opened 2025-07-16
Run program should return gas consumed in both success and failure cases

cedricjiang:


rickwebiii:
Sorry, didn't mean to close your CR

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -696,11 +694,11 @@ impl FheProcessor {
         self.aux_data.memory = None;
 
         if let Some(e) = self.aux_data.fault.get() {
-            return Err(e.clone());
+            Err(e.clone())
+        } else {
+            self.try_capture_return_value(memory, args, return_data)
+                .map(|ret_val| (gas.unwrap(), ret_val))
```
cedricjiang, line 700: unwrap is safe here

PR #132, opened 2025-07-16
Refactor gas model for more flexibility in applying different gas according to operand bit width

cedricjiang:


parasol_cpu/src/proc/gas_model.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+}
+
+impl GasModel {
+    pub(crate) fn new() -> Self {
+        let mut per_op_per_width_cost = HashMap::new();
+        let mut per_op_per_bi_width_cost = HashMap::new();
+
+        // TODO: benchmark to replace the numbers
+        per_op_per_width_cost.insert(stringify!(Not), [100_000, 100_000, 100_000, 100_000]);
```
rickwebiii, line 38: Can we put the table in assembly.rs's instruction table?

rickwebiii: You can add two values, one for plaintext and one for encrypted values

cedricjiang: also do you want gas for plaintext operations? my test indicates all plaintext operations are fast enough to be just assigned a trivial cost, though some are slightly slower than others (likely because of copying/creating ciphertexts that are relatively large in size)

cedricjiang: due to time zone differences it's hard to discuss details so let me just push this for now to make progress

PR #133, opened 2025-07-17
Use only useful bits in shift operations

cedricjiang:


mux_circuits/src/bitshift.rs
```diff
@@ -43,16 +41,14 @@ pub enum ShiftMode {
...
 /// * `dir` - The direction to shift
 /// * `mode` - The mode of the shift
 pub fn bitshift(inputs: u16, shift_size: u16, dir: ShiftDirection, mode: ShiftMode) -> MuxCircuit {
-    let used_shift_bits = ilog2_rounded_up(inputs) as u16;
-    let used_shift_bits = if used_shift_bits == 0 {
-        1
-    } else {
-        used_shift_bits
-    };
+    eprintln!("Creating circuit: {inputs}, {shift_size}");
```
cedricjiang, line 50: oops. I need to drop this out, this is for debugging

mux_circuits/src/bitshift.rs
```diff
@@ -43,16 +41,14 @@ pub enum ShiftMode {
...
+    let used_shift_bits = min_cover(inputs);
```
cedricjiang, line 51: two changes here:
1) this function just gives you the used_shift_bits, not 0 that you turn into 1
2) this function will give log2(N)+1 if N is power of 2, this makes sense because if you have input width 8, you do want to use at least 4 not 3 as 3 can only express up to 7 (this is correct previously with the "excessive bits" trick but now we do not do that in the caller which shrinks the shift_size)

mux_circuits/src/bitshift.rs
```diff
@@ -133,9 +129,10 @@ pub fn bitshift(inputs: u16, shift_size: u16, dir: ShiftDirection, mode: ShiftMo
         }
     }
 
-    // Now mux in the higher order wrapped select bits; if any are 1 then the
-    // entire output vector is zero or sign
-    if mode != ShiftMode::Rotation {
+    // Now mux in the higher order wrapped select bits if any; if any are 1 then the
+    // entire output vector is zero or sign; this does not apply to rotation as we
+    // have validated the input width is power of 2 so we can just drop higher order safely
+    if !excess_shift_vars.is_empty() && mode != ShiftMode::Rotation {
```
cedricjiang, line 135: on enter this if statement when "excess" is not empty, otherwise it's adding a few stuff while doing nothing

mux_circuits/src/bitshift.rs
```diff
@@ -310,32 +309,36 @@ mod tests {
         ] {
             for dir in [ShiftDirection::Left, ShiftDirection::Right] {
                 for width in 1..=6u16 {
-                    let mask = (1 << width) - 1;
+                    for shift_width in 1..=6u16 {
```
cedricjiang, line 312: this tests now iterates over both width and shift_width, more thorough (yet slower but ok)

PR #134, opened 2025-07-18
Support cases where flexibility is needed in handling return value with dynamic run program functions that just return unparsed bytes

cedricjiang:


cedricjiang:
I thought about different ways and they come with different drawbacks...

cedricjiang:
Ok this one seems better than previous ideas I tried

parasol_cpu/src/proc/args.rs
```diff
@@ -455,7 +455,7 @@ impl ArgsBuilder {
     /// Specify a return value for an FHE program.
     ///
     /// # Remarks
-    /// If an FHE program returns a value greater than 8 bytes, you must specify this, even if you
+    /// If an FHE program returns a value greater than 8 bytes, you must specify this or below, even if you
```
rickwebiii, line 458: This remark is now incorrect and I never updated it. You must now match the return type, full stop.

cedricjiang: do we want the comment at all then? naturally the user definitely should do this by default

parasol_cpu/src/proc/mod.rs
```diff
@@ -288,6 +288,18 @@ impl FheComputer {
         Self { processor }
     }
 
+    /// Run the given FHE program with user specified data and a gas limit, return the used gas and program raw return value
+    pub fn run_program_with_options_raw(
```
rickwebiii, line 292: `run_program_with_options_and_raw_return`

parasol_cpu/src/proc/args.rs
```diff
@@ -465,7 +465,12 @@ impl ArgsBuilder {
         }
     }
 
-    /// Specify a generic return value type for an FHE program
+    /// Specify a return value for an FHE program using the alignment and size.
+    ///
+    /// # Remarks
+    /// If an FHE program returns a value greater than 8 bytes, you must specify this or above, even if you
+    /// don't use it.
+    /// Failure to do so will result in incorrect execution.
```
rickwebiii, line 473: I don't like the `raw` nomenclature. Maybe `dynamic_return_value`?

parasol_cpu/src/proc/fhe_processor.rs
```diff
@@ -537,20 +537,20 @@ impl FheProcessor {
...
         }
     }
 
     /// Runs the given program using the passed user `data` as arguments with a gas limit
-    /// Returns the amount of gas used to run the program and the program return
-    /// value
-    pub fn run_program_with_options<T: ToArg>(
+    /// Returns the amount of gas used to run the program and the raw return
+    /// value in [`Vec<Byte>`]
+    pub fn run_program_with_options_raw(
```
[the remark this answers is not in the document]
cedricjiang: I did not "overwrite" the original `run_program_with_options` - it's right below and it calls this function and change the raw return into the type, so the the majority of the body is now in this function, leading the diff to show this as a rename

rickwebiii (approved):

parasol_cpu/src/proc/fhe_processor.rs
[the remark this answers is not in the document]
cedricjiang: open to renaming of this function if you'd like. I can also move the function so the diff looks more straightforward

PR #138, opened 2025-07-26
FFT works

rickwebiii:


sunscreen_tfhe/src/gpu/tests/fft.rs
```diff
@@ -1,28 +1,134 @@
...
+fn assert_complex_equalish<T: Float + NumCast + std::fmt::Display>(
+    actual: &Complex<T>,
+    expected: &Complex<T>,
+    eps: T,
+) {
+    let denom = if actual.re == T::from(0.0).unwrap() {
+        T::from(1.0).unwrap()
+    } else {
+        actual.re.abs()
+    };
```
ryanorendorff, line 18: This seems to be alternating between a relative and absolute epsilon for checking for equality. It may be good to do two things:

1. Provide an explicit relative and absolute epsilon tolerance
2. The code may be easier in radial (magnitude and phase) coordinates. This should also be more robust to the meaning of the complex number since checking by components is checking for magnitude and phase directly; component wise the Cartesian error leads to different achieved tolerances for on the magnitude and phase based on the angle of the complex number.

sunscreen_tfhe/src/gpu/tests/fft.rs
```diff
@@ -1,28 +1,134 @@
...
+
+    let err = (actual.im - expected.im).abs() / denom;
+
+    assert!(err < eps, "actual {actual} expected {expected}");
+}
+
+#[derive(PartialEq)]
+enum Direction {
+    Forward,
+    Reverse,
```
ryanorendorff, line 38: Inverse may make more sense here.

ryanorendorff (approved):
Looks good, only some small questions/nits

sunscreen_tfhe/src/gpu/tests/fft.rs
ryanorendorff: Alternatively you could copy what numpy does: https://github.com/numpy/numpy/blob/0532af47d6a815298b7841de00bdbc547104b237/numpy/testing/_private/utils.py#L503-L625

sunscreen_tfhe/build.rs
```diff
@@ -57,71 +57,41 @@ fn compile_as_cuda() {
...
             .arg("-o")
             .arg(&binary)
-            .arg("--verbose")
-            .args(ptx_files)
+            .arg(&src_file)
             .output()
             .unwrap();
 
-        validate_command_output(c, "nvcc linking failed.");
+        validate_command_output(c, "nvcc compilation failed.");
```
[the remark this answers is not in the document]
rickwebiii: I'll consider it for another CR. Our current compilation method requires you to have a single CUDA compilation unit, which isn't the general case. CUDA is apparently pretty bad with LTO, so having a single CU file and a bunch of headers isn't actually the worst idea, but still...

sunscreen_tfhe/src/gpu/tests/fft.rs
rickwebiii: Ideally, I'd like to express the error in number of ULPs, which is why I'm normalizing the error between 0 and 1. This should allow the caller to set a tolerance of `0.5 * op depth * f64::EPS`, but this isn't exactly a science yet. The GPU uses a sincos transcendental operation, which has 2 ULPs of error.

PR #140, opened 2025-07-29
Rweber/cuda

rickwebiii:


sunscreen_tfhe/gpu_src/src/math/fft/negacyclic.cuh
```diff
@@ -1 +1,186 @@
...
+    using S = typename ScalarOf<Complex>::Ty;
+    uint32_t n_div_2 = (S)(n >> 1);
+
+    // Split the input into 2 sub arrays, packing the first half into the real
+    // component of our twist and the second half into the imaginary part.
+    // Then, multiply by out twist factor.
+    for (uint32_t i = threadIdx.x; i < n_div_2; i += blockDim.x)
+    {   
+        // s_output[i] = twist<Complex>(i, n);
+        //s_output[i] = {(S)s_input[i], (S)s_input[i + n_div_2]};
```
rickwebiii, line 36: Remove

sunscreen_tfhe/gpu_src/src/math/fft/fft_f64_impl.cuh
```diff
@@ -23,19 +23,32 @@ SOFTWARE.
...
 
 __device__ __inline__ double2 Get_W_value_inverse_f64(int N, int m)
 {
 	double2 ctemp;
-	sincos(PI_2 * (double)m / (double)N, &ctemp.y, &ctemp.x);
+	sincos(TAU * (double)m / (double)N, &ctemp.y, &ctemp.x);
 	return ctemp;
 }
+#else
+#include "fft_constants_f64.cuh"
```
rickwebiii, line 41: Our generated constant tables currently give the wrong result. In any case, recomputing is significantly faster (~40% with f32) than a memory operation, so I'd like to avoid  this if possible. However, the accuracy of the forward twisted FFT is 1e-9, which is pretty bad. I'll revisit this accuracy issue later, but I suspect it's cascading errors with sincos().

sunscreen_gpu_runtime/src/cuda_runtime.rs
```diff
@@ -107,17 +106,19 @@ impl CudaRuntime {
     }
 }
 
-impl GpuRuntimeBackend for CudaRuntime {
-    fn print_device_info(&self, device_id: DeviceId) -> Result<()> {
+impl CudaRuntime {
+    fn set_device_id(&self, device_id: DeviceId) -> Result<()> {
```
cedricjiang, line 110: does this really need return? I don't see how the result is `Err` variant given the code

sunscreen_tfhe/src/gpu/mod.rs
```diff
@@ -1,6 +1,34 @@
...
+#[doc(hidden)]
+pub mod test_utils {
+    use std::sync::{Arc, OnceLock};
+
+    use sunscreen_gpu_runtime::GpuRuntime;
+
+    #[cfg(feature = "cuda")]
+    use sunscreen_gpu_runtime::cuda_runtime::CudaRuntime;
+
+    use super::GPU_KERNELS;
```
cedricjiang, line 18: does this const belong here?

sunscreen_gpu_runtime/src/cuda_runtime.rs
rickwebiii: I forgot to check the error code with wrap_cuda_runtime!{}.

PR #148, opened 2025-08-07
in-place polynomial fft works

rickwebiii:


sunscreen_tfhe/gpu_src/src/entities/polynomial.cuh
```diff
@@ -157,9 +159,14 @@ __device__ inline void PolynomialFft<Complex<double>>::ifft<uint64_t>(
     __syncthreads();
```
rickwebiii, line 159: Should be okay to remove this one. I think we have the guards in place to ensure you'll sync for an FFT before you go cross-lane.

PR #152, opened 2025-08-08
Implement drift-corrected noise model with rational function fitting

ryanorendorff:
Changes the error model to include the affect of the drifting encoded position of the message over the depth of the CMUX tree. This model assumes the noise spread over depth (what was characterized before), the initial encoding position distribution, and the distribution of the drift over depth are all independent. This means that the standard deviation of the estimated distribution of where a torus element could be by depth is σ²_total(d) = σ₀² + d² * σₘ² + σₛ²(d), where σ₀ is the initial encoded position distribution standard deviation, σₘ is the linear drift distribution standard deviation, and σₛ(d) is the distribution of the encoded position if no drift had occurred (the old noise estimate before this PR).

The high accuracy model can be run with

```shell
﻿﻿cargo run --features="benchmark_system_info" --release --example op_noise -- analyze-cmux-tree params --parameter-set-name default_128 --std-sample-count 10000 --std-depth 4096 --drift-sample-count 10000 --drift-depth 512
```

Resulting in this data

```
{
  "version": 2,
  "time": "2025-08-09 03:15:12.359218458 +00:00",
  "cmux_tree_parameters": {
    "parameter_set": {
      "l0_params": {
        "dim": 637,
        "std": 0.0000725
      },
      "l1_params": {
        "dim": {
          "polynomial_degree": 2048,
          "size": 1
        },
        "std": 7e-16
      },
      "cbs_radix": {
        "count": 4,
        "radix_log": 4
      },
      "pbs_radix": {
        "count": 2,
        "radix_log": 16
      },
      "ks_radix": {
        "count": 6,
        "radix_log": 2
      },
      "pfks_radix": {
        "count": 2,
        "radix_log": 17
      },
      "ss_radix": {
        "count": 15,
        "radix_log": 3
      },
      "tr_radix": {
        "count": 6,
        "radix_log": 7
      }
    },
    "run_options": {
      "drift_sample_count": 10000,
      "drift_depth": 512,
      "spread_depth": 4096,
      "spread_sample_count": 10000,
      "include_raw": false
    }
  },
  "system_info": {
    "instance_type": "c7g.16xlarge",
    "cpu_info": {
      "model_name": "Neoverse-V1",
      "cpu_cores": 64,
      "cpu_threads": 64,
      "clock_speed": 2.1,
      "low_power_mode": null,
      "avx2_support": null
    },
    "gpu_info": null,
    "ram_gb": 123
  },
  "method": "The 'random select lines, cascading data lines' method generates a set of GGSW ciphertexts for each level of the CMUX tree. At every level, two GGSWs are randomly selected—one encrypting a binary value and the other its complement—and used as select lines for CMUX operations on the data lines from the previous level. The outputs from the CMUX operations become the inputs for the next level. Noise is measured by keyswitching the resulting GLWE ciphertexts at each level into the L0 LWE key. In addition, the drift in the encoded position on the torus is calculated after each CMUX operation, which helps in understanding how the encoding drifts with each operation. Both the changes in the standard deviation and the drift helps determine the maximum CMUX tree depth that can be evaluated before bootstrapping is required to keep the probability of decryption failure below a chosen threshold (e.g., 2^-64).",
  "drift": {
    "drift_std": 0.00001879600734949663,
    "drift_offset_std": 0.004483450796344403
  },
  "fit": {
    "results": {
      "spread_model": {
        "a": 0.029052727604361776,
        "b": 2525.7137265334895,
        "c": 0.02282344595365389,
        "relative_error": 0.01962495282339516
      },
      "drift_std": 0.00001879600734949663,
      "drift_offset_std": 0.004483450796344403,
      "max_error": 0.01962495282339516,
      "base_2_error_at_depth_256": -300.39786791716034,
      "equation": "σ²_total(d) = σ₀² + d² * σₘ² + σₛ²(d), where σₛ(d) = f(x) = -1 / (a * (x + b)) + c"
    }
  }
}
```


rickwebiii (approved):
Should we move op_noise to a bin rather than an example?

ryanorendorff:
> Should we move op_noise to a bin rather than an example?

Ya probably, that would make more sense. I can do that in another PR.

PR #178, opened 2025-08-28
DynamicToArg for Vec<T: ToArg>

ryanorendorff:
Implements `DynamicToArg` for `Vec<T> where T: ToArg`. This is to enable programs to accept vectors as inputs instead of requiring arrays with known size.

### DIFF
--- a/parasol_cpu/tests/e2e_tests/vector_args.rs
+++ b/parasol_cpu/tests/e2e_tests/vector_args.rs
@@ -56,16 +56,13 @@ fn can_pass_vec_as_dynamic_arg() {
         .map(|r| r.decrypt(&enc, sk) as u8);
 
     // Expected result is element-wise addition of values_a and values_b
-    let expected: [u8; 8] = [
-        (values_a[0] + values_b[0]) as u8,
-        (values_a[1] + values_b[1]) as u8,
-        (values_a[2] + values_b[2]) as u8,
-        (values_a[3] + values_b[3]) as u8,
-        (values_a[4] + values_b[4]) as u8,
-        (values_a[5] + values_b[5]) as u8,
-        (values_a[6] + values_b[6]) as u8,
-        (values_a[7] + values_b[7]) as u8,
-    ];
+    let expected: [u8; 8] = values_a
+        .iter()
+        .zip(values_b.iter())
+        .map(|(&a, &b)| (a + b) as u8)
+        .collect::<Vec<u8>>()
+        .try_into()
+        .unwrap();
     assert_eq!(result, expected);
 }
 


cedricjiang:
Do we have a usage scenario for this?

ryanorendorff:
> Ok, I assume we have another argument (plaintext) that says how many inputs are there right?

Yes, this is how most of the C programs are set up that take arrays, regardless of if they are arrays or vecs.

cedricjiang:
Yeah, unless the function is designed to know how many elements

parasol_cpu/src/proc/args.rs
```diff
@@ -409,6 +409,65 @@ impl DynamicToArg for DynamicInt<L1GlweCiphertext> {
     }
 }
 
+impl<T: ToArg> DynamicToArg for Vec<T> {
+    fn alignment(&self) -> usize {
+        T::alignment()
+    }
```
cedricjiang, line 415: the `alignment(&self)`

parasol_cpu/src/proc/args.rs
```diff
@@ -555,4 +619,143 @@ mod tests {
...
+
+        // u8 has no padding (alignment = 1)
+        let vec_u8: Vec<u8> = vec![1, 2, 3];
+        assert_eq!(vec_u8.size(), 3);
+
+        // u16 has alignment = 2, but since size = 2, no extra padding needed
+        let vec_u16: Vec<u16> = vec![1, 2, 3];
+        assert_eq!(vec_u16.size(), 6); // 3 * 2
+
+        // For custom types with different size/alignment, padding would matter
```
cedricjiang, line 741: would like to see an example for this

parasol_cpu/src/proc/args.rs
ryanorendorff: This matches the `ToArg` definition for an array. We are defining how large the elements are, not the size here. 

See https://github.com/Sunscreen-tech/spf/blob/fe17efed940ce6ef84d539ee49b48f073a7333f6/parasol_cpu/src/proc/args.rs#L265-L267

parasol_cpu/src/proc/args.rs
```diff
@@ -409,6 +409,65 @@ impl DynamicToArg for DynamicInt<L1GlweCiphertext> {
...
+        }
+
+        self.iter()
+            .flat_map(|x| {
+                // Pad each vector element to its alignment (same pattern as arrays)
+                let bytes = x.to_bytes();
+
+                bytes
+                    .iter()
+                    .chain(std::iter::repeat(&bytes[0]))
```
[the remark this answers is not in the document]
cedricjiang: I see. you can pad anything so whatever is convenient

parasol_cpu/src/proc/args.rs
```diff
@@ -409,6 +409,65 @@ impl DynamicToArg for DynamicInt<L1GlweCiphertext> {
...
+
+        let padded_size = T::size().next_multiple_of(T::alignment());
+
+        if data.len() % padded_size != 0 {
+            return Err(Error::TypeSizeMismatch);
+        }
+
+        let elements = data
+            .chunks(padded_size)
+            .map(|chunk| T::try_from_bytes(chunk.to_vec()))
```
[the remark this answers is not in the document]
ryanorendorff: Removed, now uses the same implementation as the array.

parasol_cpu/src/proc/args.rs
ryanorendorff: No more padding, no more tests.

PR #197, opened 2025-09-14
Rweber/addends

rickwebiii:


sunscreen_tfhe/src/math/simd/generic/mod.rs
```diff
@@ -124,6 +124,16 @@ pub fn complex_mad(c: &mut [Complex<f64>], a: &[Complex<f64>], b: &[Complex<f64>
     scalar::complex_mad(c, a, b)
 }
 
+/// Compute vector `c += a * b` over &[Complex<f64>].
```
cedricjiang, line 127: I guess it's `-=` as the function name includes `sub`?

PR #219, opened 2025-12-11
Simple malloc instruction support

cedricjiang:


parasol_cpu/src/proc/gas_model.rs
```diff
@@ -144,7 +144,7 @@ impl GasModel {
             // instructions that have trivial cost: either they do not deal with any ciphertext at all
             // or they don't compute on ciphertext content (just treat them as vector of opaque objects)
             Load(..) | LoadI(..) | Store(..) | BranchNonZero(..) | BranchZero(..) | Branch(..)
-            | Move(..) | Dbg(..) | Sext(..) | Zext(..) | Trunc(..) => 1,
+            | Move(..) | Dbg(..) | Sext(..) | Zext(..) | Trunc(..) | Malloc(..) => 1,
```
cedricjiang, line 147: I suppose this is fair...

parasol_cpu/src/proc/ops/malloc.rs
```diff
@@ -0,0 +1,56 @@
...
+            match size {
+                Register::Plaintext { val, width } => {
+                    if *width != 32 || *val >= u32::MAX as u128 {
+                        return Err(Error::IllegalOperands {
+                            inst_id: instruction_id,
+                            pc,
+                        });
+                    }
+
+                    let ptr = memory.try_allocate(*val as u32)?;
```
cedricjiang, line 32: this allocates page boundary so alignment is not a problem, but this is quite wasteful...

I am not sure if we want to implement some actual allocation mechanism at this point

rickwebiii: This isn't actually that wasteful. This will allocate 4kB of `Byte`s, which default to plaintext. Between the enum discriminant and payload, you're looking at like 8kB I think, which is less than a single encrypted bit.

rickwebiii (approved):

parasol_cpu/src/proc/ops/malloc.rs
cedricjiang: that's true...

cedricjiang: one thing to note is this wastes "address space", let's say it's ok for now assuming we don't have a lot of mallocs by the user

PR #221, opened 2026-03-19
Rweber/greco

rickwebiii:


sunscreen_tfhe/src/math/torus.rs
```diff
@@ -50,6 +52,29 @@ where
...
 
+/// Generate a random element of this type.
+pub trait Random {
+    /// Generate a uniform element of this type.
+    fn uniform() -> Self;
+
+    /// Generate an element from a normal distribution characterized by the given [Stddev].
+    fn normal<F>(std: Stddev, sampler: F) -> Self
+    where
+        F: FnOnce(&Normal<f64>) -> f64;
```
cedricjiang, line 63: I wonder if this f64 can be a generic type...

sunscreen_tfhe/src/math/torus.rs
```diff
@@ -198,6 +223,82 @@ where
...
+        let q = (u32::BITS as f64).exp2();
+        let e = f64::round(sample * q) as i32;
+        let e: u32 = i32::cast_unsigned(e);
+        u32::from_u64(e as u64)
+    }
+
+    fn binary() -> Self {
+        Self::uniform() & 0x1
+    }
+}
```
cedricjiang, line 270: these two impls are close enough to use macro, just saying

probably same for below

sunscreen_tfhe/src/math/torus.rs
```diff
@@ -198,6 +223,82 @@ where
...
+
+impl Encoding for u64 {
+    fn decode(&self, plaintext_bits: PlaintextBits) -> Self {
+        let round_bit = self.wrapping_shr(Self::BITS - plaintext_bits.0 - 1) & Self::from_u64(0x1);
```
cedricjiang, line 274: u64::from_u64 :)

sunscreen_tfhe/src/zkp.rs
cedricjiang, line 1: didn't realize this file existed...

cedricjiang: ok it's a full rewrite...

sunscreen_tfhe/src/math/fgreco.rs
```diff
@@ -0,0 +1,500 @@
...
+struct BarrettConfig;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(transparent)]
+/// The finite field used in the Greco proof system, which allows generating proofs of encryption.
+pub struct FGreco(Zq<1, BarrettBackend<1, BarrettConfig>>);
+
+impl ToF64 for FGreco {
+    fn to_f64(self) -> f64 {
+        unimplemented!()
```
cedricjiang, line 37: todo or not intended to be implemented? if latter, should explicit panic with a message saying this is not supposed to be called at all

same for later ones in the file

sunscreen_tfhe/src/math/fgreco.rs
```diff
@@ -0,0 +1,500 @@
...
+use crate::{
+    Encoding, FromF64, FromU64, NumBits, Random, ReinterpretAsSigned, ToF64, ToU64, TorusOps,
+    simd::VectorOps,
+};
+
+/// The modulus used for the field used by Greco.
+pub const GRECO_MODULUS: u64 = 6943179709095039;
+
+#[derive(BarrettConfig)]
+#[barrett_config(modulus = "6943179709095039", num_limbs = 1)]
```
cedricjiang, line 27: I wonder if we can consolidate this string const and the u64 const above

cedricjiang: this is not 2^n, but I assume the math in the file is correct

rickwebiii: No, `BarretConfig` derive macro looks for a string literal after modulus. This is a limitation of codegen, so there isn't a way I can think of to consolidate these.

rickwebiii: > this is not 2^n, but I assume the math in the file is correct

I'm using an existing macro I wrote for generating Barrett reductions over rings. That code is tested and we used it with SDLP.

sunscreen_tfhe/src/math/fgreco.rs
rickwebiii: Unimplemented implies "You shouldn't call this because it's never going to be implemented."

rickwebiii: The string you get if you try is

> thread 'main' (13) panicked at src/main.rs:2:5:
not implemented

which I find sufficient for debugging.

sunscreen_tfhe/src/math/torus.rs
rickwebiii: Yes, but there's no reason to generate these at lower precision.

sunscreen_tfhe/src/math/torus.rs
rickwebiii: Won't fix

sunscreen_tfhe/src/math/torus.rs
rickwebiii: I prefer to use `Self` in all impls.

sunscreen_tfhe/src/math/torus.rs
cedricjiang: yeah, not worth it

sunscreen_tfhe/src/math/torus.rs
cedricjiang: I don't mean you should replace `Self::from_u64` with `u64::from_u64`, I mean it's kind of funny that the code is converting from a type to itself, so technically you can forget about it and just do `& 0x1` at the end.

again no need to make changes

